PR #19, opened 2016-11-17
Feature fix build case sensitive os

talosdev:


talosdev:
Wrong base branch (master instead of develop)

PR #21, opened 2016-11-29
Custom logger

talosdev:
Replace LogCat with SLF4J and logback to generate log files for debugging


smartReceiptsPRO/build-types/release/assets/logback.xml
```diff
@@ -0,0 +1,29 @@
...
+        <tagEncoder>
+            <pattern>%logger{12}</pattern>
+        </tagEncoder>
+        <encoder>
+            <pattern>[%date] &lt;%logger{20}&gt; %level: %msg%n</pattern>
+        </encoder>
+    </appender>
+
+    <appender name="file" class="ch.qos.logback.core.rolling.RollingFileAppender">
+        <file>/data/data/wb.receiptspro/files/log.txt</file>
```
wbaumann: My only question is with respect to this line (for all logback.xml files but only calling out here). 

Do you know if Android changes these paths on different devices? For example, context#getExternalFilesDir(null) can return each of the following based on the manufacturer:

- /sdcard/Android/data/wb.receipts/files
- /storage/emulated/0/Android/data/wb.receipts/files
- /storage/ABCD-0123/Android/data/wb.receipts/files
- ...

As these are all defined by the manufacturer. I can also test in the wild but was curious if logback supports a wildcard here like {files-dir} or something?

wbaumann:
Going to pull this in now to avoid some upcoming merge conflicts (as we can always clean up later and potentially address the note above in an upcoming fix). Let me know!

smartReceiptsPRO/build-types/release/assets/logback.xml
talosdev: That's a quite valid observation. It seems that the internal storage path is more standardized that the external one, but in the documentation there is a warning that the path might change in future releases of the platform.
There seems to be a way to use predefined-variables or custom ones in the xml file. I will try out if this works, and that way I could also reduce the 4 different xml conf files to just one.

https://github.com/tony19/logback-android/issues/64

wbaumann: Cool. Your reference link looks very promising, so hopefully we can just use that (and reduce the number of XML files too!). Let me know how it works out

PR #22, opened 2016-11-30
Feature custom logger

talosdev:
Improvements in logback configuration xml. 
Log directory path is configured through code.
Better handling (with includes) for switching the logcat appender on/off depending on whether the build is debug/release.


wbaumann:
Looks great. Thanks for the clean up!

PR #25, opened 2016-12-07
Feature tooltip customize reports

talosdev:


smartReceiptsLibrary/src/main/res/values/strings.xml
```diff
@@ -248,6 +248,8 @@
     <string name="DIALOG_EMAIL_NEGATIVE_BUTTON">Cancel</string>
     <string name="DIALOG_EMAIL_TOAST_NO_SELECTION">Please Check At Least One Report</string>
     <string name="DIALOG_EMAIL_TOAST_NO_RECEIPTS">There are No Receipts in this Report</string>
+    <string name="DIALOG_EMAIL_TOOLTIP">Configure your report columns and layout</string>
```
wbaumann: Hey in the future, let's keep these all lower case. I kept some of the original ones all upper case (for legacy reasons), but I want to align with Android standards in the future. No need to change here though

PR #26, opened 2016-12-09
Feature admob native ads

talosdev:
I ended up providing different sizes for the banner based on two qualifiers, screen height and orientation, because otherwise it didn't look good on landscape.


wbaumann:
Thanks for this. Will let you know if there are any subscription issues from my end but looks to be all good!

PR #33, opened 2017-03-20
Integration with Dagger2

JuliaSoboleva:
I came to such modules for now:

- BaseAppModule. Provides general things which are same for Free and Pro versions
- GlobalBindingModule. Makes all subcomponents (== all activities and fragments that need some injection) blinded to the AppComponent. https://google.github.io/dagger/android.html
- FreeAppModule and ProAppModule. Provides special things like AdManager and PurchaseWallet.

Components and subcomponents:

- FreeAppComponent and ProAppComponent. Main components that use (Free/Pro)AppModule, BaseAppModule and GlobalBindingModule. I think with flavors we will have just one AppComponent which will use needed flavor's AppModule.
- Subcomponents for each activity and fragment that needs some injection.

Public methods that are already removed from the SmartReceiptsApplication class:

- getWorkerManager()
- getPersistenceManager()
- getServiceManager()
- getFlex()
- getFleXML()
- instantiateWorkerManager(), instantiatePersistenceManager(), instantiateFlex()

Btw, I've also added a couple of protected methods (getPersistenceManagerInternal() and getReceiptColumnDefinitionsInternal()), but I think they will be removed after migration to the flavors structure.

Looking forward to your review :)


smartReceiptsFree/src/main/java/wb/receipts/di/FreeAppComponent.java
```diff
@@ -0,0 +1,54 @@
...
+            return application;
+        }
+
+        @Provides
+        @ApplicationScope
+        public static AdManager provideAdManager(FreeAdManager freeAdManager) {
+            return freeAdManager;
+        }
+
+        @Provides
```
wbaumann: Any reason this isn't tied to the @ApplicationScope? I think this class has singleton-like characteristics, since it is managed via SharedPrefs, but it may be easier that way

smartReceiptsLibrary/src/main/AndroidManifest.xml
```diff
@@ -5,6 +5,7 @@
     android:versionCode="1"
     android:versionName="1.0" >
 
+    <uses-permission android:name="android.permission.WAKE_LOCK" />
```
wbaumann: Does dagger need WAKE_LOCK? Or can we get rid of this?

smartReceiptsLibrary/src/main/java/co/smartreceipts/android/SmartReceiptsApplication.java
```diff
@@ -118,10 +114,8 @@ private void configureLog() {
     public void onTerminate() {
```
wbaumann: Feel free to delete this entire method if it's causing much work, since we don't really leverage it for UTs

smartReceiptsFree/src/main/java/wb/receipts/ad/FreeAdManager.java
```diff
@@ -35,8 +19,25 @@
...
+import co.smartreceipts.android.purchases.PurchaseableSubscriptions;
+import co.smartreceipts.android.purchases.Subscription;
+import co.smartreceipts.android.purchases.SubscriptionEventsListener;
+import co.smartreceipts.android.purchases.source.PurchaseSource;
+import co.smartreceipts.android.purchases.wallet.PurchaseWallet;
+import co.smartreceipts.android.utils.log.Logger;
+import wb.receipts.R;
+
+@ApplicationScope
+public class FreeAdManager implements AdManager, SubscriptionEventsListener {
```
wbaumann: Just to note, I have a ticket open to refactor this more gracefully, since this is pretty messy

smartReceiptsLibrary/src/main/java/co/smartreceipts/android/di/BaseAppModule.java
```diff
@@ -0,0 +1,42 @@
...
+    public static Flex provideFlex(Context context) {
+        return new Flex(context, new Flexable() {
+            @Override
+            public int getFleXML() {
+                return Flexable.UNDEFINED;
+            }
+        });
+    }
+
+    @Provides
```
wbaumann: @ApplicationScope here?

smartReceiptsLibrary/src/main/java/co/smartreceipts/android/di/BaseAppModule.java
```diff
@@ -0,0 +1,42 @@
...
+    public static StorageManager provideStorageManager(Context context) {
+        return StorageManager.getInstance(context);
+    }
+
+    @Provides
```
wbaumann: @ApplicationScope here?

smartReceiptsLibrary/src/main/java/co/smartreceipts/android/persistence/PersistenceManager.java
```diff
@@ -1,81 +1,81 @@
...
+import co.smartreceipts.android.di.scopes.ApplicationScope;
 import co.smartreceipts.android.settings.UserPreferenceManager;
 import wb.android.storage.InternalStorageManager;
 import wb.android.storage.SDCardFileManager;
 import wb.android.storage.SDCardStateException;
 import wb.android.storage.StorageManager;
-import co.smartreceipts.android.SmartReceiptsApplication;
 
+@ApplicationScope
 public class PersistenceManager {
```
wbaumann: As we go along, feel free to remove this class as well if it's simpler to just use the @Inject calls for all of it's subclasses (eg StorageManager, DatabaseHelper, etc).

smartReceiptsLibrary/src/main/java/co/smartreceipts/android/settings/widget/ColumnsListFragment.java
```diff
@@ -174,9 +175,9 @@ public void onDeleteFailure(@NonNull Column<Receipt> column, @Nullable Throwable
         mDisableInsertsAndDeletes = false;
     }
 
-    protected ColumnDefinitions<Receipt> getColumnDefinitions() {
-        return new ReceiptColumnDefinitions(getActivity(), getPersistenceManager(), getFlex());
-    }
+//    protected ColumnDefinitions<Receipt> getColumnDefinitions() {
```
wbaumann: Can we delete this now? Looks like the @Inject works for subclasses?

smartReceiptsLibrary/src/main/AndroidManifest.xml
JuliaSoboleva: Sorry, just my mistake. I've let Android Studio add this permission because of some code snippet which was marked as lint error.

smartReceiptsFree/src/main/java/wb/receipts/di/FreeAppComponent.java
JuliaSoboleva: You are right. I'll add @ApplicationScope for PurchaseWallet

smartReceiptsLibrary/src/main/java/co/smartreceipts/android/di/BaseAppModule.java
JuliaSoboleva: Great catch! I haven't added this scope because I've looked on it like on a part of PersistenceManager which has @ApplicationScope. But I've missed other usages of StorageManager.
So, I'll add @ApplicationScope and injections for other classes that use StorageManager.getInstance() method.

smartReceiptsLibrary/src/main/java/co/smartreceipts/android/settings/widget/ColumnsListFragment.java
JuliaSoboleva: All subclasses (PDFColumnsListFragment and CSVColumnsListFragment) already have injected ReceiptColumnDefinitions. But superclass has some logic in onCreate method which initializes mSpinnerAdapter and needs ReceiptColumnDefinitions.
Now we have protected abstract getReceiptColumnDefinitions() method which is implemented in subclasses and allows to pass injected ReceiptColumnDefinitions.

smartReceiptsLibrary/src/main/java/co/smartreceipts/android/di/BaseAppModule.java
JuliaSoboleva: As about DatabaseHelper, I didn't add @ApplicationScope for it because it has some custom logic in getInstance() method. New object creates not just when INSTANCE == null. It also creates when INSTANCE.isOpen() is false. 
So, I decided to leave this logic as is and don't assign it to Dagger.

wbaumann: Cool makes sense!

smartReceiptsLibrary/src/main/java/co/smartreceipts/android/settings/widget/ColumnsListFragment.java
wbaumann: Perfect. Thanks for all the updates!

PR #34, opened 2017-03-24
Feature flavors

JuliaSoboleva:
Two flavors: **free** and **pro**.

Now we have just one `SmartReceiptsApplication`, so we can inject dependencies directly and there is no need for redundant methods like `getSmthInternal()`.

As about dagger's modules and components, now we have single `AppComponent`, which binds needed `AppModule` from flavor's directory.

Also, I've added `ExtraInitializer` which is used for specific flavor-dependent init process. In fact `ExtraInitializerPlusImpl` does nothing and `ExtraInitializerFreeImpl` inits BugsenseHandler and registers GoogleAnalytics.

One more small fix. I noticed that in the last pull request I've broken CalendarDialog because of injecting application context instead of activity context. Now it's fixed.


app/build.gradle
```diff
@@ -0,0 +1,121 @@
...
+        targetSdkVersion rootProject.ext.targetSdkVersion
+        multiDexEnabled true
+        vectorDrawables.useSupportLibrary = true
+    }
+
+    productFlavors {
+        free {
+            applicationId "wb.receipts"
+        }
+        pro {
```
wbaumann: Just as an fyi - I'm going to merge this in and handle a few rebase conflicts along the way.

As I do, I'm going to rename Pro -> Plus, since I've changed all the user facing strings as well

JuliaSoboleva: Thanks for your participation!

wbaumann:
Rebased all this in. Broke a couple UTs in the InteractivePdf stack, so I'll address those as part of the next commit. Thanks!!

PR #35, opened 2017-03-31
No more getters in the application class

JuliaSoboleva:
Hey Will,

Looks like we got rid of all getters from the application class, so take a look.

I notice that it's easier to move forward with Dagger after a fairly slow start. Hope you like how the code changes :)

Some detail note: I've optimized a bit `TableControllerManager`'s getters usages and made all managers which are contained by it injectable. Thus, now there is just one class that uses all those getters - `GoogleDriveBackupManager`. 
We can go forward and сompletely get rid of getters and inject all directly in `GoogleDriveBackupManager`. And then possibly the "technical" need for the `TableControllerManager` class will be gone. But on the part of logic and readability of the code, `TableControllerManager` makes sense and grouping special managers together. 
So, what way do you think is better: leave it as is with getters which are used by one `GoogleDriveBackupManager` class; get rid of all getters but leave `TableControllerManager` class, or get rid of getters and `TableControllerManager` class?
I personally think the best is the second option, some kind of a compromise, but I need your advice.

Thanks and have a great day!


app/src/free/java/co/smartreceipts/android/ad/FreeAdManager.java
```diff
@@ -45,15 +43,21 @@
...
     private PurchaseManager purchaseManager;
 
+
     @Inject
-    FreeAdManager(@NonNull PurchaseWallet purchaseWallet) {
-        this.purchaseWallet = Preconditions.checkNotNull(purchaseWallet);
+    FreeAdManager(PurchaseWallet purchaseWallet, Analytics analytics) {
+        // note: Dagger throws exception if there is null in parameters.
+        // So, no need in additional check
+//        this.purchaseWallet = Preconditions.checkNotNull(purchaseWallet);
```
wbaumann: Perfect - now that dagger is handling the nulls for us, feel free to delete these commented out lines

app/src/free/java/co/smartreceipts/android/di/AppModule.java
```diff
@@ -30,4 +38,11 @@ public static PurchaseWallet providePurchaseWallet (DefaultPurchaseWallet defaul
     public static ExtraInitializer provideExtraInitializer (ExtraInitializerFreeImpl freeInitializer) {
         return freeInitializer;
     }
+
+    @Provides
+    @ApplicationScope
+    public static Analytics provideAnalytics(FirebaseAnalytics firebaseAnalytics, GoogleAnalytics googleAnalytics) {
```
wbaumann: With the double injects, does this work for the Plus version, which doesn't have GoogleAnalytics?

app/src/main/java/co/smartreceipts/android/activities/SmartReceiptsActivity.java
```diff
@@ -266,6 +270,6 @@ public void run() {
 
     @Nullable
     public PurchaseManager getSubscriptionManager() {
-        return mPurchaseManager;
+        return purchaseManager;
```
wbaumann: Note (also to myself), but this Activity getter is fairly useless now that we have Dagger

app/src/main/java/co/smartreceipts/android/apis/hosts/ServiceManager.java
```diff
@@ -16,6 +17,7 @@
  * Provides a standardized manner in which we can define host configurations and their association to a specific
  * service endpoint for network requests within the app
  */
+@ApplicationScope
```
wbaumann: What's the purpose to adding @ApplicationScope to these and the others as opposed to in the @Provides module location?

app/src/main/java/co/smartreceipts/android/di/BaseAppModule.java
```diff
@@ -54,4 +67,36 @@ public static DatabaseHelper provideDatabaseHelper(Context context, StorageManag
...
+    }
+
+    @Provides
+    @ApplicationScope
+    public static AppRatingStorage provideAppRatingStorage(AppRatingPreferencesStorage storage) {
+        return storage;
+    }
+
+    @Provides
+    @co.smartreceipts.android.di.qualifiers.ReceiptColumnDefinitions
```
wbaumann: Custom Subcomponent? Or is this for something else?

app/src/plusflavor/java/co/smartreceipts/android/di/AppModule.java
```diff
@@ -30,4 +37,11 @@ public static PurchaseWallet providePurchaseWallet (PlusPurchaseWallet plusPurch
     public static ExtraInitializer provideExtraInitializer (ExtraInitializerPlusImpl plusInitializer) {
         return plusInitializer;
     }
+
+    @Provides
+    @ApplicationScope
+    public static Analytics provideAnalytics(FirebaseAnalytics firebaseAnalytics) {
```
wbaumann: Ah nvm my comment above - I see how you're handling this :)

app/src/main/java/co/smartreceipts/android/activities/SmartReceiptsActivity.java
JuliaSoboleva: You are absolutely right, I'll remove it in next commit :)

app/src/main/java/co/smartreceipts/android/apis/hosts/ServiceManager.java
JuliaSoboleva: It's not necessary but I duplicate scopes before class declaration based on reasons that most of injectable classes have no (and need no) @Provides at the module. Thus, most of injectable classes if they have any scope - have it specified above the class declaration. Just some kind of unification.

app/src/main/java/co/smartreceipts/android/di/BaseAppModule.java
JuliaSoboleva: It's a custom qualifier (could be replaced with standard @Named("some_name") annotation). I've added it because of hypothetical possibility of injecting some `ColumnDefinitions<Receipt>` which is not `ReceiptColumnDefinitions`.
Actually, it's not necessary at this stage.

wbaumann: Cool makes complete sense. Thanks for clarifying!

app/src/main/java/co/smartreceipts/android/apis/hosts/ServiceManager.java
wbaumann: Gotcha. Thanks for letting me know! I'll follow the same pattern when adding new items to Dagger2

PR #36, opened 2017-04-06
FragmentScope applying

JuliaSoboleva:
Hi Will,

I've changed `TableControllerManager`to `GoogleDriveTableManager `as you offered. Now it looks really better :)

Also, I've worked on @FragmentScope applying. 

Each subcomponent of the injectable fragment has this scope, and using of @FragmentScope for some object inside of fragment class means "local singleton". Thus, based on MVP model, the presenter is a perfect candidate for this scope. But now we have just a few fragment's presenters, so you can see @FragmentScope using just for `TripFragmentPresenter`, `TripCreateEditFragmentPresenter`, and `ReceiptCreateEditFragmentPresenter`. 

Others fragments have no presenters or have presenter and interactor that are closely related to views. So, I think we can leave the task of Dagger integration and continue to apply it throughout the next tasks and some future refactoring. Do you agree with me?


wbaumann:
Looks really great :)

And I agree! Lets' leave the task for future Dagger integration as part of future work. These commits take care of the majority of it, so it'll be relatively trivial to get set up for the next stage! As we go along too, we'll probably want to look into even narrower scopes (ie instead of @FragmentScope, we'll have @TripDisplayScope - here's a decent example: https://github.com/twistedequations/android-rxmvp/tree/master/app/src/main/java/com/twistedequations/rxmvp), but I think this is ideal for the first pass. 

Thanks again for all of this!

PR #38, opened 2017-04-20
Upgrade to Rx2 and Retrolambda

JuliaSoboleva:
Looks like it's finally done :)

I've removed Rx1 dependency and also updated Retrofit version to completely get rid of the old Rx dependency in the dependency tree.

Where it was obvious, I tried to replace `Observable` with `Single`, `Completable` and `Maybe` types. But despite this, I'm sure there are still some places where such replacement is appropriate.

As about tests, I have the same result as upstream code version from `develop` branch (I have 40 tests failed, but some of them are related to localization - like expected string is "20.2" and actual is "20,2" ). So, I hope I haven't broken anything.

Thanks


wbaumann:
Ah good note about the localization issues! I'll have to go ahead and try to correct these and issue another fix for them

app/src/main/java/co/smartreceipts/android/aws/cognito/CognitoIdentityProvider.java
```diff
@@ -35,74 +31,62 @@ public CognitoIdentityProvider(@NonNull IdentityManager identityManager, @NonNul
...
-                    public Cognito call(Throwable throwable) {
-                        return null;
-                    }
-                })
-                .toBlocking().value();
+        try {
+            return refreshCognitoToken()
+                    .onErrorReturn(throwable -> Optional.absent())
+                    .blockingGet().get();
+        } catch (Exception e) {
```
wbaumann: Just curious - do we need the catch clause here or is the "onErrorReturn" clause sufficient?

app/src/main/java/co/smartreceipts/android/imports/ActivityFileResultImporter.java
```diff
@@ -75,50 +73,49 @@ public ActivityFileResultImporter(@NonNull Context context, @NonNull FragmentMan
...
-                        analytics.record(new ErrorEvent(ActivityFileResultImporter.this, throwable));
+                    public void onComplete() {
+                        headlessFragment.importSubject.onComplete();
                     }
-                })
-                .observeOn(observeOnScheduler)
-                .subscribe(headlessFragment.importSubject);
+                });
+
+        // TODO: 12.04.2017 !!! check dispose(). not sure if it works correctly
```
wbaumann: Todo note - still relevant?

app/src/main/java/co/smartreceipts/android/persistence/database/controllers/TableController.java
```diff
@@ -51,15 +52,14 @@
      *
      * @param modelType the object to insert
      * @param databaseOperationMetadata metadata about this particular database operation
-     * @return a hot {@link Observable} that will return the result of this operation
+//     * @return a hot {@link Observable} that will return the result of this operation
      */
-    @NonNull
-    Observable<ModelType> insert(@NonNull ModelType modelType, @NonNull DatabaseOperationMetadata databaseOperationMetadata);
+    /*Observable<ModelType>*/void insert(@NonNull ModelType modelType, @NonNull DatabaseOperationMetadata databaseOperationMetadata);
```
wbaumann: Nobody is using this at the moment (so no need to change in the immediate term), but my goal is to allow the callers of #insert(...) to receive an Observable with the exact result (or onError)

Down stream consumers can use the #getInsertStream(...) to get a full list of results

The basic challenge here is that some entities need to know about a single insert (eg the UI) whereas others need to know about all of them (eg auto-backups). I hadn't figured out a good way to handle this otherwise, since input takes in parameters (and you couldn't find the observable without knowing the input params in advance). Let me know if you've seen any better patterns here.

app/src/main/java/co/smartreceipts/android/aws/cognito/CognitoIdentityProvider.java
JuliaSoboleva: You're right, it's unnecessary. And we can pass null through the onErrorReturn.

app/src/main/java/co/smartreceipts/android/imports/ActivityFileResultImporter.java
JuliaSoboleva: No, it's no relevant already. Missed it

PR #39, opened 2017-04-27
Small fix for bug with back navigation after new/edit trip save

JuliaSoboleva:
Little fix and long text :)

When we tried to save new/edited trip with `TripCreateEditFragment`, we had back stack with such transactions: 

- replace(`TripFragment`, `TripCreateEditFragment`) == remove(`TripFragment`), add(`TripCreateEditFragment`)
- replace(`TripCreateEditFragment`, `ReportInfoFragment`) == remove(`TripCreateEditFragment`), add(`ReportInfoFragment`)

So, when we click back button, `onBackPressed()` is called -> `popBackStackImmediate()` -> calling last transaction reverced -> remove(`ReportInfoFragment`), add(`TripCreateEditFragment`).

At first glance behavior that we need for `TripCreateEditFragment` looks similar to `ReceiptCreateEditFragment`, but when `ReceiptCreateEditFragment` is opened we have such back stack:
- replace(`TripFragment`, `ReportInfoFragment`)
- replace(`ReportInfoFragment`, `ReceiptCreateEditFragment`)

And `ReportInfoFragment` is that one which will be added when the back button will be pressed.

Thus, with `TripFragment`->`TripCreateEditFragment`->`ReportInfoFragment `chain we need to "forget" about `TripCreateEditFragment`.

I solved it by one extra calling of `popBackStackImmediate()`, so when we click save button on `TripCreateEditFragment`, last transaction from back stack is popped and then `ReportInfoFragment` is opened. Back stack works like this:
- replace(`TripFragment`, `TripCreateEditFragment`)
- `popBackStackImmediate() `
- replace(`TripFragment`, `ReportInfoFragment`).

Now `onBackPressed()` does correct things and `TripCreateEditFragment` is "forgotten".

This solution makes barely visible blink, but it's much less tricky than other possible solutions.

### DIFF
--- a/app/src/main/java/co/smartreceipts/android/activities/NavigationHandler.java
+++ b/app/src/main/java/co/smartreceipts/android/activities/NavigationHandler.java
@@ -79,6 +79,11 @@ public class NavigationHandler {
         }
     }
 
+    public void navigateToReportInfoFragmentWithoutBackStack(@NonNull Trip trip) {
+        mFragmentManager.popBackStackImmediate();
+        navigateToReportInfoFragment(trip);
+    }
+
     public void navigateToCreateNewReceiptFragment(@NonNull Trip trip, @Nullable File file, @Nullable OcrResponse ocrResponse) {
         if (mIsDualPane) {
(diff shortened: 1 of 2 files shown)


wbaumann:
Thanks for the detailed description. Looks good :)

PR #40, opened 2017-05-07
Bugfix TransactionTooLargeException

JuliaSoboleva:
`NavigationHandler` now is injectable (so I've removed `DaggerFragmentNavigationHandler`). I created `SmartReceiptsActivityBindingModule` and moved some fragment's subcomponents from `GlobalBindingModule`, which allows them to inject `NavigationHandler` (because it has `ActivityScope`).

Now we have injectable `FragmentStateCache` and we can put/get arguments/state `Bundle` and prevent `TransactionTooLargeException`. I hope :)

Some notes about `FragmentStateCache`'s methods:
- `putArguments` should be used before calling fragment's newInstance();
- `putSavedState` should be used in fragment's onSaveInstanceState;
- `onDestroy` should be used in fragment's onDestroy to prevent a memory leak.


wbaumann:
Thank you so much! This looks really great, and I think moving the FragmentInjector to the Activity class will be especially helpful going forward. 

Can you double check that the following flow still works:

1. Settings
2. Customize CSV Reports

Since I know this also requires a fragment (presumably from the SmartReceiptsApplication fragment injector?). 

Great work and thank you!

JuliaSoboleva:
Yes, I think Customize CSV Reports still works. 
New changes did not affect this part, `SettingsActivity` and `SettingsViewerActivity` don't need anything with `@ActivityScope`, and all fragments from `SettingsViewerActivity` are created with empty `newInstance()` method

PR #41, opened 2017-05-11
Fix for Rx undeliverable exception 

JuliaSoboleva:
I've found crush caused by Rx undeliverable exception when the user flow is: menu -> Automatic Scans (OCR) -> back (without internet connection).

Looks like `emitter` is already disposed but we try to call `emitter.onError` and Rx throws that undeliverable exception.
Thus, I've wrapped this call with an additional check and it doesn't crush more.


wbaumann:
Thanks Julia! Appreciate the find and fix!

PR #42, opened 2017-05-16
Improve ads code

JuliaSoboleva:
Now we have `SmartRecieptsActivityAdModule`s for free and plus flavors that provides `AdPresenter` (`EmptyBannerAdPresenter` for plus and `NativeBannerAdPresenter` or `ClassicBannerAdPresenter` for free version - depends on boolean constant `USE_CLASSIC_ADS`).

Upsell button now is embedded into `BannerAdView` (interface which is implemented by `ClassicAdView` and `NativeAdView`)

Ad Unit Id for classic ads I added as a string resource with name `classicAdUnitId`.

And I think it's all main changes :)


app/src/free/java/co/smartreceipts/android/di/SmartReceiptsActivityAdModule.java
```diff
@@ -0,0 +1,24 @@
...
+import co.smartreceipts.android.ad.impl.ClassicBannerAdPresenter;
+import co.smartreceipts.android.ad.impl.NativeBannerAdPresenter;
+import co.smartreceipts.android.di.scopes.ActivityScope;
+import dagger.Module;
+import dagger.Provides;
+
+@Module
+public class SmartReceiptsActivityAdModule {
+
+    private final static boolean USE_CLASSIC_ADS = true;
```
wbaumann: Could you add this to `FeatureFlags.java`, so we can track this all in one place

wbaumann:
Looks great! I'm going to pull this in as is. Could you at some point move the `USE_CLASSIC_ADS` Boolean to `FeatureFlags.java`, so we can track all these in a single place?

Also does this handle the XML error in in `app/src/free/res/layout/activity_main_twopain.xml`:

`<include layout="@layout/ad_layout" ...`

I wanted to ensure these are corrected at some point if possible. We can handle this via Android Studio > Anaylize > Inspect Code...

Overall - looks great though and thank you for your time :)

PR #43, opened 2017-05-17
Added UseNativeAds(false) to FeatureFlags

JuliaSoboleva:
I've added `UseNativeAds` to `FeatureFlags`.

As about lint error relating to `ad_layout`, at first I didn't fully understand the problem, but the thing is now we have no such errors because all code that somehow uses ads now is located in the free flavor module (I mean all `AdPresenter` implementations except `EmptyBannerAdPresenter` which is in the plus flavor).


wbaumann:
Looks great. Thanks!

PR #44, opened 2017-05-21
Injectable ActivityFileResultImporter

JuliaSoboleva:
Now `ActivityFileResultImporter` is injectable, has `@ApplicationScope` and doesn't use headless fragment.
`FileImportProcessorFactory` now is also injectable (without scope).

Generally, there are no serious changes, just some kind of refactoring :)


wbaumann:
Thanks, @JuliaSoboleva! Looks great

PR #45, opened 2017-06-03
Improve tooltip usage

JuliaSoboleva:
It's finally done :)

It took longer than I expected. I tried to repeat class structure that you use for `LoginFragment`-`LoginPresenter`-`LoginInteractror`.

Thus, now we have:
- `ReportTooltipFragment`: a fragment which can show report-related tooltips, i.e. two types of tooltips - different sync errors or generate info which is aiming to help user to find 'Generate' tab;
- `ReportTooltipPresenter`: extends `BasePresenter`;
- `ReportTooltipInteractor`.

And some additional classes:
- `GenerateInfoTooltipPreferencesStorage`: shared preferences wrapper to save if generate info tooltip was ever dismissed or if the user generated report
- `GenerateInfoTooltipManager`: decides whether there is a need to show generate info tooltip. It takes into account the condition of `GenerateInfoTooltipPreferencesStorage` and checks the number of user's trips. If the user has a single trip, it checks the number of receipts.
- `ReportTooltipUiIndicator`: contains a state of the tooltip (`SyncError` or `GenerateInfo`, or `None`) and optional `SyncErrorType`.
- `GenerateNavigator`: interface for a class that is able to move to 'Generate' tab - `ReportInfoFragment` in our case.

Also, I've covered with tests interactor and presenter. But I had to test GenerateInfoTooltipManager manually because it's pretty tricky to mock `DatabaseHelper` behavior.

Take a look when you will have time :)
Thanks


app/src/main/java/co/smartreceipts/android/widget/tooltip/report/ReportTooltipFragment.java
```diff
@@ -0,0 +1,130 @@
...
+    private void presentGenerateInfo() {
+        tooltip.setInfo(R.string.tooltip_generate_info_message,
+                v -> {
+                    tooltipClickStream.onNext(ReportTooltipUiIndicator.generateInfo());
+
+                    Fragment parentFragment = this.getParentFragment();
+                    if (parentFragment instanceof GenerateNavigator) {
+                        ((GenerateNavigator) parentFragment).navigateToGenerateTab();
+                    } else {
+                        throw new IllegalStateException("Parent fragment must implement GenerateNavigator interface");
```
wbaumann: Let's move this check into `onCreate` so we fail fast here

app/src/main/java/co/smartreceipts/android/widget/tooltip/report/generate/GenerateInfoTooltipManager.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @Inject
+    public GenerateInfoTooltipManager(DatabaseHelper databaseHelper, GenerateInfoTooltipStorage storage) {
+        this.databaseHelper = databaseHelper;
+        this.preferencesStorage = storage;
+    }
+
+    public Single<Boolean> needToShowGenerateTooltip() {
+
+        return databaseHelper.getTripsTable()
```
wbaumann: To our point of this being difficult to mock, I completely agree. A goal of mine is to slowly get everything into the TableController implementation in a fully mockable manner (eg `TripTableController#getStream())

PR #46, opened 2017-06-29
Feature graphs

JuliaSoboleva:
New features:
- added two new options for Plus flavor: "Separate payments by Category" (adds to full PDF and CSV report separate table for each category) and "Include categorical summation" (ads to full PDF and CSV report one new table with sum expenses for categories);

- added new "Graphs" tab to ReportInfoFragment at the 0-position. For now there are available 4 charts:

1. LineChart for daily expenses;
2. PieChart for expenses grouped by category;
3. Horizontal stacked BarChart for user's reimburseable vs non-reimburseable expenses;
4. BarChart for user's expenditures by payment method.

First two charts are a bit interactive, others are static.

As for chart's value format, all prices are converted to the currency that is default for the current trip, but I decided not to show currency symbol so that signatures do not look cumbersome. But we can change it if it's not obvious :)


app/src/main/java/co/smartreceipts/android/graphs/GraphsPresenter.java
```diff
@@ -0,0 +1,61 @@
...
+    public void subscribe(Trip trip) {
+        this.trip = Preconditions.checkNotNull(trip);
+
+        subscribe();
+    }
+
+    @Override
+    public void subscribe() {
+
+        // TODO: 21.06.2017 we have some nuance with payment methods.
```
wbaumann: Is TODO still relevant or just a note?

app/src/main/java/co/smartreceipts/android/graphs/entry/LabeledGraphEntry.java
```diff
@@ -0,0 +1,42 @@
...
+    public LabeledGraphEntry(float value, String label) {
+        super(value, label);
+    }
+
+    public String getLabel() {
+        return (String)super.getData();
+    }
+
+    @Override
+    public boolean equals(Object obj) {
```
wbaumann: We need to override hashcode here too to make this work, correct? Or is this what IntelliJ produced for it's generation?

app/src/main/java/co/smartreceipts/android/persistence/database/controllers/grouping/results/CategoryGroupingResult.java
```diff
@@ -0,0 +1,26 @@
...
+import co.smartreceipts.android.model.Receipt;
+
+public class CategoryGroupingResult {
+
+    private final Category category;
+
+    private final List<Receipt> receipts;
+
+    public CategoryGroupingResult(Category category, List<Receipt> receipts) {
+        this.category = category;
```
wbaumann: No worries for this one but in the future for some of these that we create, can you use the @NonNull annotation along with:

this.category =  Preconditions.checkNotNull(category)

To verify that nobody breaks the nullability contract. Thanks!

app/src/main/java/co/smartreceipts/android/persistence/database/controllers/grouping/results/CategoryGroupingResult.java
```diff
@@ -0,0 +1,26 @@
...
+        this.receipts = receipts;
+    }
+
```
wbaumann: Plus @NonNull for the return types. I'm only noting it for this one but it's helpful for all model objects to take advantage of IntelliJs built-in nullability analysis

app/src/test/java/co/smartreceipts/android/graphs/GraphsInteractorShould.java
```diff
@@ -0,0 +1,231 @@
...
+import co.smartreceipts.android.persistence.database.controllers.grouping.GroupingController;
+import io.reactivex.Single;
+
+import static co.smartreceipts.android.R.string.graphs_label_non_reimbursable;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.when;
+
+@RunWith(RobolectricTestRunner.class)
+public class GraphsInteractorShould {
```
wbaumann: In the future (no worries here), could you change the suffix from "Should" to "Test". I actually like "Should" better as a way of indicating the test :), but I don't think AndroidStudio/IntelliJ picks this up by default when you use the CMD+Shift+T test launch short cut. Unless this is configurable if you know?

app/src/main/java/co/smartreceipts/android/graphs/GraphsPresenter.java
```diff
@@ -0,0 +1,61 @@
...
+
+import co.smartreceipts.android.di.scopes.FragmentScope;
+import co.smartreceipts.android.model.Trip;
+import co.smartreceipts.android.persistence.DatabaseHelper;
+import co.smartreceipts.android.settings.UserPreferenceManager;
+import co.smartreceipts.android.settings.catalog.UserPreference;
+import co.smartreceipts.android.widget.viper.BasePresenter;
+
+@FragmentScope
+public class GraphsPresenter extends BasePresenter<GraphsView, GraphsInteractor> {
```
wbaumann: Could you add UT coverage for this one too to ensure that it's relaying the interactor logic around to our View?

JuliaSoboleva: Sure, will do

app/src/main/java/co/smartreceipts/android/graphs/GraphsPresenter.java
JuliaSoboleva: Yes, I need to clarify with you this moment.

For now, if Payment Methods disabled (as default) all is ok. Let's say I have 5 receipts. And then I turn on using of Payment Methods at the Settings page - for old receipts and for Graphs nothing will be changed because old receipts have no payment method.

Then if I add new receipt, it has default payment method - "Unspecified". And I will see this receipt at Graphs tab, as expected.

But isn't it a bit confusing? I have 6 receipts, but at the chart I see only one with "Unspecified" payment method. 

Maybe it would be more proper to make PaymentMethod field @NonNull and set "Unspecified" payment method for all new receipts even if Payment Methods are disabled at Settings.
What do you think about it?

wbaumann: @JuliaSoboleva l think that's a great idea! The only challenge here (and probably reason for not having this in the first place) is that the "Unspecified" payment method is editable in the menu (hence allowing users to alter/edit/delete this), which would make things trickier. 

Maybe we should do this in conjunction with locking edits to that item? Or creating a new "Unspecified" one? Let me open up a new Trello task for this, and we can move the discussion there. Thanks!

app/src/main/java/co/smartreceipts/android/graphs/entry/LabeledGraphEntry.java
JuliaSoboleva: Yep, I lose it maybe because Android Studio couldn't generate such method due to lack of own fields for this class. 
I'll add something like:

```
@Override
    public int hashCode() {
        return getData().hashCode();
    }
```

app/src/test/java/co/smartreceipts/android/graphs/GraphsInteractorShould.java
JuliaSoboleva: I liked to use "Should" in favor of the verbal combination: class name + name of the method.
But you're right, let it be more unified along the project :)

PR #47, opened 2017-07-06
Fixes for graphs

JuliaSoboleva:
Small fixes for Graphs

### DIFF
--- a/app/src/main/java/co/smartreceipts/android/graphs/DayAxisValueFormatter.java
+++ b/app/src/main/java/co/smartreceipts/android/graphs/DayAxisValueFormatter.java
@@ -4,9 +4,9 @@ import com.github.mikephil.charting.components.AxisBase;
 import com.github.mikephil.charting.formatter.IAxisValueFormatter;
 
 import java.util.Calendar;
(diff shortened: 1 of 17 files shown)


app/src/main/java/co/smartreceipts/android/graphs/DayAxisValueFormatter.java
```diff
@@ -26,7 +25,11 @@ public String getFormattedValue(float value, AxisBase axis) {
             int day = calendar.get(Calendar.DAY_OF_MONTH);
             int month = calendar.get(Calendar.MONTH) + 1;
 
-            return String.format("%02d", day) + defaultSeparator + String.format("%02d", month);
+            if (Locale.getDefault() == Locale.US) {
```
wbaumann: I still think this won't be sufficient, since some other countries also occasionally do MM/DD instead of DD/MM.

We probably need to use something like `DateFormat.java` for this: https://stackoverflow.com/questions/18480633/java-util-date-format-conversion-yyyy-mm-dd-to-mm-dd-yyyy

PR #54, opened 2017-07-16
NonNull payment methods

JuliaSoboleva:
Now Payment Methods must be @NonNull and we have a new one `ImmutablePaymentMethodImpl.NONE` which is default for all new Receipts.

It would be pretty hard to use a custom string for this one due to lack of `Context` and I decided to use Android system string `<Untitled>`. I think it's quite suitable for our case :)

New Payment Method is visible for user at the payment methods spinner at position 0 from the `ReceiptCreateEditFragment` but it's not present at the payment methods editor. Also receipts with `ImmutablePaymentMethodImpl.NONE` payment method are not counted in the payment methods chart.

Maybe now we can remove `Unspecified` payment method from defaults? :)


wbaumann:
Awesome. Thank you :)

PR #57, opened 2017-08-09
PDF validation

JuliaSoboleva:
I've added a small check for image import (directly in `ImageImportProcessor.process` body) and `PdfValidator` class which allows us to check if the pdf file can be read. 

Actually, "wrong" image file (for ex which is a text file in fact) couldn't be imported even without this changes, but now we have a special `InvalidImageException` for this case.

All import operations need `READ_EXTERNAL_STORAGE` permission, but we had no this check before and import process failed with a general error message but without any explanations for the user. So I've added permission check when the user clicks on the `Import` button. If we have no `READ_EXTERNAL_STORAGE` permission, the app will ask the user to give it. If he declines our request we will show him toast with `"Cannot import file without storage permissions"` message. 
One strange note: we had a string with such toast message but it wasn't used and wasn't translated...

For permission checking, I've added `PermissionsDelegate` class which actually just combines work of  `PermissionStatusChecker` and `PermissionRequester` and returns simple `Completable`. Pretty сonvenient tool to use I think :)

And about testing. I really tried to test not valid files but it's some sort of pain. For example, when Roboelectric tries to read image file which is not actually an image, roughly speaking
 I expect to get -1 values of image's width and height (like real Android does) or even 0, but I get 100 because of `ShadowBitmapFactory` internal work   ¯ \ _ (ツ) _ / ¯

### DIFF
--- a/app/src/main/java/co/smartreceipts/android/imports/ActivityFileResultImporter.java
+++ b/app/src/main/java/co/smartreceipts/android/imports/ActivityFileResultImporter.java
@@ -1,6 +1,8 @@
 package co.smartreceipts.android.imports;
 
+import android.Manifest;
 import android.app.Activity;
+import android.content.ContentResolver;
 import android.content.Intent;
 import android.net.Uri;
 import android.support.annotation.NonNull;
@@ -15,13 +17,15 @@ import javax.inject.Inject;
 
 import co.smartreceipts.android.analytics.Analytics;
 import co.smartreceipts.android.analytics.events.ErrorEvent;
-import co.smartreceipts.android.di.scopes.ApplicationScope;
+import co.smartreceipts.android.di.scopes.ActivityScope;
 import co.smartreceipts.android.model.Trip;
 import co.smartreceipts.android.ocr.OcrManager;
+import co.smartreceipts.android.permissions.PermissionsDelegate;
 import co.smartreceipts.android.utils.log.Logger;
 import io.reactivex.Maybe;
 import io.reactivex.Observable;
 import io.reactivex.Scheduler;
+import io.reactivex.Single;
 import io.reactivex.android.schedulers.AndroidSchedulers;
 import io.reactivex.disposables.Disposable;
 import io.reactivex.observers.DisposableObserver;
@@ -29,12 +33,13 @@ import io.reactivex.schedulers.Schedulers;
 import io.reactivex.subjects.ReplaySubject;
 import io.reactivex.subjects.Subject;
 
-@ApplicationScope
+@ActivityScope
 public class ActivityFileResultImporter {
 
     private final Analytics analytics;
     private final OcrManager ocrManager;
     private final FileImportProcessorFactory factory;
+    private final PermissionsDelegate permissionsDelegate;
 
     private final Scheduler subscribeOnScheduler;
     private final Scheduler observeOnScheduler;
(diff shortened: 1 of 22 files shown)


app/src/main/java/co/smartreceipts/android/imports/utils/PdfValidator.java
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    public boolean isPdfValid(@NonNull File file) {
+        PdfPDImageXFactory factory = new PdfPDImageXFactoryFactory(context, new PDDocument(), file).get();
+        try {
+            factory.open();
+            // document has at least one page == it's valid
+            return factory.nextPage();
+        } catch (IOException e) {
+            e.printStackTrace();
```
wbaumann: Let's change this to Logger.error(this, "Invalid PDF File", e);

as part of a future CR

app/src/main/java/co/smartreceipts/android/permissions/PermissionsDelegate.java
```diff
@@ -0,0 +1,41 @@
...
+    @Inject
+    PermissionStatusChecker permissionStatusChecker;
+    @Inject
+    ActivityPermissionsRequester<SmartReceiptsActivity> permissionRequester;
+
+    @Inject
+    public PermissionsDelegate() {
+    }
+
+    public Completable checkPermissionAndMaybeAsk(@NonNull String manifestPermission) {
```
wbaumann: Nice work :). Very convenient!

app/src/main/java/co/smartreceipts/android/receipts/ReceiptsListFragment.java
```diff
@@ -299,8 +304,21 @@ public final void addTextReceipt() {
     }
 
     private void importReceipt() {
-        final ImportPhotoPdfDialogFragment fragment = new ImportPhotoPdfDialogFragment();
-        fragment.show(getChildFragmentManager(), ImportPhotoPdfDialogFragment.TAG);
+        permissionsDelegate.checkPermissionAndMaybeAsk(Manifest.permission.READ_EXTERNAL_STORAGE)
```
wbaumann: I'm not sure that this piece is required here. Or rather, we should only need it if the file we attempt to import is on Android M and above and uses a file scheme (ie `file://...`). If it uses a content schema (ie `content://`), we shouldn't require this.

Would it be possible for us to only prompt for permissions after the fact if the user is trying to import a `file://` schemed object? I have a similar flow in place for whenever the user opens an image file outside of Smart Receipts and clicks the send/share icon to attach it to Smart Receipts. Check out `IntentImportInformationInteractor.java`'s `process(Intent)` method to see how it handles these different types.

JuliaSoboleva:
@wbaumann I've fixed permission check behavior. Didn't know about this details about reading files before :)

app/src/main/java/co/smartreceipts/android/imports/ActivityFileResultImporter.java
```diff
@@ -15,26 +17,29 @@
...
+import io.reactivex.Single;
 import io.reactivex.android.schedulers.AndroidSchedulers;
 import io.reactivex.disposables.Disposable;
 import io.reactivex.observers.DisposableObserver;
 import io.reactivex.schedulers.Schedulers;
 import io.reactivex.subjects.ReplaySubject;
 import io.reactivex.subjects.Subject;
 
-@ApplicationScope
+@ActivityScope
```
wbaumann: I think this should be good but now that we're changing from `ApplicationScope` to `ActivityScope`, could you test the following (no rush though):

1. Enable `"Don't keep activities"` in the developer settings menu of your Android device
2. Add a `delay(10, TimeUnit.Seconds)` to this `ActivityFileResultImporter` stream
3. Rotate the device half way through the import

I just want to confirm that the import isn't "lost" during this circumstance

PR #58, opened 2017-08-26
ActivityFileResultImporter with ApplicationScope again

JuliaSoboleva:
It was a challenging part and I don't like how the code looks like now, but now import process is pretty stable during the device rotation.

There used to be such algorithm: `ActivityFileResultImorter` has `AppplicationScope` and inner `ReplaySubject`. Fragment subscribes to subject when `onResume` is called and dispose `ActivityFileResultImorter` when the import is done. 

What `ActivityFileResultImorter` does:
1. `getSaveLocation` - returns `Uri` if all ok
2. `process(uri)` - actual import

Now we need to wedge between these steps a scheme check and if the scheme is `file` check `READ_EXTERNAL_STORAGE` permission and asking it if needed.

Permission check is an activity scoped action and I tried to combine it with  `getSaveLocation`  step because  `getSaveLocation` is quick. But we're getting new problems with permission dialog because while it's being shown user also can rotate device and resubscriptions in such state make a lot of mess.

I tried different options but came to one which works.
For now, the import process is a chain of actions:
1. `ActivityFileResultLocator` does `getSaveLocation` from the first importer step
2. At the fragment code we check and ask permission if needed
3. `ActivityFileResultImportet` does actual import

'ActivityFileResultLocator' and 'ActivityFileResultImportet' have `ApplicationScope`. To save the state when rotating  and handle possible errors, we subscribe to locator and importer not at `onResume` but at `onCreate` and also `setReatinInstance` set to `true` (for example if user denied permission we need to show some toast message about permission but if we will resubscribe at `onResume` user will never know why nothing happened). Locator's and Importer's inner subject are `PublishSubject`s.

I can't say that this way is the best but I didn't find another working one.


app/src/main/java/co/smartreceipts/android/fragments/ReceiptImageFragment.java
```diff
@@ -170,13 +169,48 @@ public void onActivityResult(final int requestCode, final int resultCode, final
...
+                        progress.setVisibility(View.GONE);
+                    }
+                }));
+    }
+
+    private void dispose() {
+        activityFileResultImporter.dispose();
+        activityFileResultLocator.dispose();
+
+        compositeDisposable.dispose();
```
wbaumann: FYI - You can just call `compositeDisposable#clear()`, so you don't have to `null` it out at the end. This combines disposing everything and clearing out the underlying list of `Disposable`s.

app/src/main/java/co/smartreceipts/android/receipts/ReceiptsListFragment.java
```diff
@@ -198,36 +208,72 @@ public void onActivityCreated(Bundle savedInstanceState) {
...
+                    } else {
+                        Toast.makeText(getActivity(), getFlexString(R.string.FILE_SAVE_ERROR), Toast.LENGTH_SHORT).show();
+                        highlightedReceipt = null;
+                        updatingDataProgress.setVisibility(View.GONE);
                     }
-                }, throwable -> {
-                    Toast.makeText(getActivity(), getFlexString(R.string.FILE_SAVE_ERROR), Toast.LENGTH_SHORT).show();
-                    highlightedReceipt = null;
-                    updatingDataProgress.setVisibility(View.GONE);
-                    activityFileResultImporter.dispose();
```
wbaumann: One question/thought from the old approach. It exhibited the following flow (which as you pointed out is different by necessity as we need to go from `ApplicationScope` -> `ActivityScope` (for permissions) -> `ApplicationScope`):

* Kick off the locate/import process via the `ActivityFileResultsImporter`, which was tied to the `ApplicationScope`
* Should the user rotate the device during this time-frame, we leverage a `ReplaySubject`, so the act of re-subscribing will always return us the previous stream of events (including `onError` or `onComplete` ones).
* Hence upon seeing the `onError` or `onComplete` event, we know that we're "done" for the first time and inform our parent object to tear down it's stream (ie the call to `activityFileResultImporter.dispose();` here)

Did you run into any issues with something similar? I think your idea here is the best one (ie `Locator` -> Permissions -> `Importer`), but I was wondering if we could use a `ReplaySubject` or `BehaviorSubject` instead of a `PublishSubject`, so we don't need to use the `setRetainInstance(true)` call. 

Thoughts - I wasn't sure if you came across something else here that prevented this?

One other trick I've found in this area (since I assume you may have avoided the `onError` and `onComplete` calls by wrapping everything with the `getThrowable().isPresent()` stuff in `onNext`) is to use a `BehaviorSubject` in order to always get last event. Then you can do something like:

```
return behaviorSubject.filter(Optional::isPresent)
```

And whenever the user calls `dispose` or something to indicate that "I consumed both Locate/Import calls", you can just do:

```
behaviorSubject.onNext(Optional.absent())
```

And hence subscribers will no longer see the "last" event when rotating again (as it'll be filtered out). To apply this, you can just call `locator#dispose` and `import#dispose` after receiving a `importer -> onNext`response rather than doing it in the `dispose` method in both fragments.

The `RxInAppBillingServiceConnection.java` class is kind of an example of this approach.

JuliaSoboleva: Thanks for sharing your ideas!

I worked a bit on it and looks like I can remove `setReatinInstance(true)` and use `BehaviourSubject` with the trick that you described :)

But we still must subscribe at `onCreate` and dispose at `onDestroy`. For example, when permission dialog is shown and the user clicks on Reject button we can't catch this event because our fragment's `onResume` will be called, we will subscribe and locator's `BehaviourSubject` will return the previous result again, and permission dialog will be shown again. I mean in such snippet:

```
compositeDisposable.add(activityFileResultLocator.getUriStream()
                .flatMapSingle(response -> {
                    if (response.getUri().getScheme().equals(ContentResolver.SCHEME_CONTENT)) {
                        return Single.just(response);
                    } else { // we need to check read external storage permission
                        return permissionsDelegate.checkPermissionAndMaybeAsk(READ_PERMISSION)
                                .toSingleDefault(response)
// we will never come to next
                                .doOnError(throwable -> {
                                    Toast.makeText(getActivity(), "ERROR", Toast.LENGTH_SHORT).show();
                                    Logger.debug(this, "Importer permission denied");
                                })
                                .onErrorReturn(ActivityFileResultLocatorResponse::LocatorError);
                    }
                })
```
we will never come to `permissionsDelegate`'s errors handling.

I'm testing changed approach now. I'll clean it a bit, apply it for another `ReceiptImageFragment` and I'll add new commit here.

Thanks!

wbaumann: @JuliaSoboleva Sounds good and makes a ton of sense about subscribing in `onCreate` to properly handle the dialog button clicks. Thanks for looking into this!

wbaumann:
@JuliaSoboleva Awesome work! Thank you so much for taking the time and testing through all these cases. I know that Android lifecycle is messy and difficult, but I think you've found a great solution!

JuliaSoboleva:
@wbaumann My code solution would have been worse without your advice. Thank you :)

PR #61, opened 2017-11-01
Drag & drop for Categories and Payment Methods

JuliaSoboleva:
Hi Will,

This task is very long, but I think it's worth it :)

The version of the DB is increased. Now `Categories`, `PaymentMethods`, `Receipts` (for the future), `CSVColumns` and `PDFColumns` tables have new `custom_order_id` columns. Default value is 0. Also, `Categories` table is refactored and now we have `id` column as a primary key.

New `OrderingPreferencesManager` stores boolean flags for any table with reordering option, for example, `ORDERING_CATEGORIES` flag. 

Thus, user receives categories in default order (for example ordered by name for categories - like it was) before he opens categories editor fragment. After he saw editor regardless of whether he changed anything, all categories will receive their sequential number, which will be saved to DB as `custom_order_id`'s. Next time user will receive categories ordered by `custom_order_id` column.

My next step is adding reordering mode for CSV and PDF columns editor fragment.

What do you think about reordering for receipts? Should we do it after CSV and PDF columns reordering implementation or better to leave it for later?

BTW, I really like new app icon changes, looks nicer :)


app/src/main/java/co/smartreceipts/android/model/Draggable.java
```diff
@@ -0,0 +1,9 @@
+package co.smartreceipts.android.model;
+
+public interface Draggable {
```
wbaumann: Does this always get used with the `Comparable<T>` implementations? If so, let's have it extend `Comparable<T>`

app/src/main/java/co/smartreceipts/android/model/impl/ImmutableCategoryImpl.java
```diff
@@ -11,65 +11,88 @@
...
+        return syncState;
+    }
+
+    @Override
+    public int getCustomOrderId() {
+        return customOrderId;
     }
 
     @Override
     public boolean equals(Object o) {
```
wbaumann: Unrelated to your changes, but I eventually want to move our Model Objects to be Kotlin data classes, since these will automatically update our `equals` and `hashCode` implementations. 

We can also use https://github.com/google/auto/tree/master/value if we stick with Java, but I think this will reduce future code churn :)

app/src/main/java/co/smartreceipts/android/model/impl/ImmutableCategoryImpl.java
```diff
@@ -98,9 +123,8 @@ public ImmutableCategoryImpl createFromParcel(Parcel source) {
 
     };
 
-
     @Override
     public int compareTo(@NonNull Category category) {
-        return mName.compareTo(category.getName());
+        return customOrderId - category.getCustomOrderId();
```
wbaumann: Can we add a fallback here if both have a `customOrderId = 0` to use the `name` field instead? This assumes that I was correct in understanding that we don't set this value until the first time the user starts to sort them

app/src/main/java/co/smartreceipts/android/persistence/DatabaseHelper.java
```diff
@@ -512,19 +516,19 @@ public final synchronized boolean merge(String dbPath, String packageName, boole
...
+                                if (packageName.equalsIgnoreCase("wb.receipts")) {
+                                    newParent = oldParent.replace("wb.receiptspro/", "wb.receipts/");
+                                } else if (packageName.equalsIgnoreCase("wb.receiptspro")) {
+                                    newParent = oldParent.replace("wb.receipts/", "wb.receiptspro/");
                                     }
                                     File f = new File(newParent);
                                     newParent = f.getName();
                                 }
-                                final String category = getString(c, categoryIndex, "");
+                            final int categoryId = getInt(c, categoryIdIndex, 0);
```
wbaumann: Can we add a fallback here to better support cases in which a user is merging in an old database? Something like:

```
final String category = getString(c, categoryIndex, "");
final int categoryId = getInt(c, categoryIdIndex, -1);
if (categoryId > 0) {
    values.put(ReceiptsTable.COLUMN_CATEGORY_ID, categoryId);
} else {
    values.put(ReceiptsTable.COLUMN_CATEGORY, category);
}
```

I'm not sure if this would work, but I want to ensure that the categories still link up for databases before v16.

app/src/main/java/co/smartreceipts/android/persistence/database/defaults/TableDefaultCustomizerImpl.java
```diff
@@ -24,13 +24,11 @@
 
 public class TableDefaultCustomizerImpl implements TableDefaultsCustomizer {
 
-    private final Context context;
```
wbaumann: Not a big deal, but let's try to use the `private final` approaches in general as opposed to using injection outside of the constructor

app/src/main/java/co/smartreceipts/android/persistence/database/tables/ReceiptsTable.java
```diff
@@ -209,6 +210,78 @@ public synchronized void onUpgrade(@NonNull SQLiteDatabase db, int oldVersion, i
...
+            db.execSQL(addCategoryIdColumn);
+
+            final String fillCategoryId = String.format("UPDATE %s SET %s = ( SELECT %s FROM %s WHERE %s = %s LIMIT 1 )",
+                    ReceiptsTable.TABLE_NAME, COLUMN_CATEGORY_ID, CategoriesTable.COLUMN_ID, CategoriesTable.TABLE_NAME,
+                    CategoriesTable.COLUMN_NAME, OLD_COLUMN_CATEGORY);
+            Logger.debug(this, fillCategoryId);
+            db.execSQL(fillCategoryId);
+
+            // removing category's name column
+            final String copyTable = "CREATE TABLE " + ReceiptsTable.TABLE_NAME + "_copy" + " ("
```
wbaumann: I can't recall how everything is setup, but I'm assuming that `CategoryTable#onUpgrade` is always called before `ReceiptsTable#onUpgrade`? Since we'll need to reference it's new key?

We may also wish to add the include transaction inside a start/end/success block:

```
db.beginTransaction();
try {
    // For each #onUpgrade
    db.setTransactionSuccessful();
} catch {
    //Error in between database transaction 
} finally {
    db.endTransaction();
}
```

This way we're safe if anything fails halfway through and nothing gets corrupted
@see https://stackoverflow.com/questions/8147440/android-database-transaction

app/src/main/java/co/smartreceipts/android/persistence/database/tables/adapters/CategoryDatabaseAdapter.java
```diff
@@ -47,6 +53,9 @@ public ContentValues write(@NonNull Category category, @NonNull DatabaseOperatio
         final ContentValues values = new ContentValues();
         values.put(CategoriesTable.COLUMN_NAME, category.getName());
         values.put(CategoriesTable.COLUMN_CODE, category.getCode());
+//        if (category.getCustomOrderId() != CategoryBuilderFactory.MISSING_ID) {
```
wbaumann: Delete?

app/src/main/java/co/smartreceipts/android/persistence/database/tables/adapters/ReceiptDatabaseAdapter.java
```diff
@@ -125,7 +130,7 @@ public Receipt readForSelection(@NonNull Cursor cursor, @NonNull Trip trip, bool
         final ReceiptBuilderFactory builder = new ReceiptBuilderFactory(id);
         builder.setTrip(trip)
                 .setName(name)
-                .setCategory(categoryImpl)
+                .setCategory(categoryOptional.orNull())
```
wbaumann: Right now `Receipt.getCategory` is set to return @NonNull. Is there a another way to handle this? Or do we need to make it @nullable.

app/src/main/java/co/smartreceipts/android/persistence/database/tables/ordering/OrderingPreferencesManager.java
```diff
@@ -0,0 +1,66 @@
...
+
+import javax.inject.Inject;
+
+import co.smartreceipts.android.di.scopes.ApplicationScope;
+
+@ApplicationScope
+public class OrderingPreferencesManager {
+
+    @Inject
+    Context appContext;
```
wbaumann: Let's make this final and move it to the constructor. Plus easier to Unit Test :)

app/src/test/java/co/smartreceipts/android/persistence/database/tables/ReceiptsTableTest.java
```diff
@@ -361,6 +361,18 @@ public void onUpgradeFromV14() {
         verifyV14Upgrade(times(1));
     }
 
+//    @Test
+//    public void onUpgradeFromV15() {
```
wbaumann: Is it possible to test this one? Or too complex?

app/build.gradle
```diff
@@ -101,6 +101,10 @@ android {
 
         compile "com.github.PhilJay:MPAndroidChart:$MP_ANDROID_CHART_VERSION"
 
+        compile ("com.h6ah4i.android.widget.advrecyclerview:advrecyclerview:$ADV_RECYCLER_VIEW_VERSION@aar"){
```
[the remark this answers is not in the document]
JuliaSoboleva: Done :)
One moment. For example, we have `support-v4` dependency for both `wbMiniLibrary` module and `app` module. What do you think we should do? use `api` for `wbMiniLibrary` and remove it from `app` or maybe duplicates look more clearly?

app/src/main/java/co/smartreceipts/android/model/Draggable.java
JuliaSoboleva: Agreed and done.

app/src/main/java/co/smartreceipts/android/model/impl/ImmutableCategoryImpl.java
JuliaSoboleva: Looking forward to work with Kotlin :)

app/src/main/java/co/smartreceipts/android/model/impl/ImmutableCategoryImpl.java
JuliaSoboleva: As I found, compare is used for sorting cashed data list when there are some changes occurred. For example, for editing item - updated item will be removed from cash and added to the end of the list. Then cashed list will be sorted. 
Thus, I'm not sure if this case when all `customOrderId`'s are 0 is possible but I've added this check for categories because of alphabetic order.  However, payment methods had no such order preferences so I think it makes sense to leave everything as is for it.

app/src/main/java/co/smartreceipts/android/persistence/DatabaseHelper.java
JuliaSoboleva: I've missed merging process, will work on it.

When we are doing merge we can have imported DB with any version. I just thought maybe we can call `onUpgrade` first to have current and imported DS's with the same version and then make all merge operation? Because, for example, old `Receipts` table had `CATEGORY` column with the category name, but new `Receipts` table has no such column, just `CATEGORY_KEY` and would be good to upgrade old table first (and category name will be changed to category id). What do you think about it?

Also, I think, I found a little bug. When the app doesn't have read permission and we are trying to import from within the app some backup file - the app doesn't ask permission and just shows `"Error: Failed to recognize the file as importable"`. But if I try to open `*.smr` file from outside using the app - all is ok, the app asks read permission. Looks like we need to add permission check

app/src/main/java/co/smartreceipts/android/persistence/database/defaults/TableDefaultCustomizerImpl.java
JuliaSoboleva: Accepted. I'll now use this approach :)

app/src/main/java/co/smartreceipts/android/persistence/database/tables/ReceiptsTable.java
JuliaSoboleva: Yes, `ReceiptsTable` is upgraded after all others because of:

```
mTables.add(mTripsTable);
        mTables.add(mDistanceTable);
        mTables.add(mCategoriesTable);
        ....
        mTables.add(mReceiptsTable);

for (final Table table : mTables) {
            table.onUpgrade(db, oldVersion, newVersion, mCustomizations);
        }
```

I'll read more about transactions and implement it :)

app/src/main/java/co/smartreceipts/android/persistence/database/tables/adapters/CategoryDatabaseAdapter.java
JuliaSoboleva: Yep :)

app/src/main/java/co/smartreceipts/android/persistence/database/tables/adapters/ReceiptDatabaseAdapter.java
JuliaSoboleva: You are right. I changed `setCategory` and `setPaymentMethod` methods to take @nonnull argument.

app/src/main/java/co/smartreceipts/android/persistence/database/tables/ordering/OrderingPreferencesManager.java
JuliaSoboleva: Done

app/src/test/java/co/smartreceipts/android/persistence/database/tables/ReceiptsTableTest.java
JuliaSoboleva: Yes, now I've added a test similar to what already exists for this class.

app/build.gradle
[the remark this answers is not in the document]
wbaumann: `api` is fine there. I think we'll probably want to remove `wbMiniLibrary` at some point, since I don't use it for any other projects :). I'll take care of that later on

app/src/main/java/co/smartreceipts/android/persistence/DatabaseHelper.java
wbaumann: @JuliaSoboleva Great idea. I think running through all the `#onUpgrade` scripts would make it far easier than my current approach.

And good catch with the bug. Could you open a new Trello task for that in the `Bugs` list? That way we can track this going forward. This change already has enough included, so I figured we'd save that for another one

app/src/main/java/co/smartreceipts/android/persistence/database/tables/ReceiptsTable.java
wbaumann: Perfect! And the transactions stuff just makes sure that things unwind if it ever fails. I think we can just do:

```
db.beginTransaction();
try {
    for (final Table table : mTables) {
        table.onUpgrade(db, oldVersion, newVersion, mCustomizations);
    }
    db.setTransactionSuccessful();
} catch {
    //Error in between database transaction 
} finally {
    db.endTransaction();
}
```

wbaumann:
@JuliaSoboleva I hope I didn't break anything... I had made some changes recently to add `constraintLayout` support, which I added here:

https://github.com/wbaumann/SmartReceiptsLibrary/pull/61/commits/8cff7bc8983072426de9980a913beacad978a28d

Did this come through okay on your end? I can merge the rest if so.

app/src/main/java/co/smartreceipts/android/persistence/DatabaseHelper.java
JuliaSoboleva: Ok, I've added new Trello [task](https://trello.com/c/h4ykajz7/286-no-read-permission-request-during-the-importing-backup-file-from-within-the-app)

app/src/main/java/co/smartreceipts/android/persistence/database/tables/ReceiptsTable.java
JuliaSoboleva: In `SQLiteOpenHelper#onUpgrage` [documentation]( https://developer.android.com/reference/android/database/sqlite/SQLiteOpenHelper.html#onUpgrade(android.database.sqlite.SQLiteDatabase,%20int,%20int) ) I've found this line 
`"This method executes within a transaction. If an exception is thrown, all changes will automatically be rolled back."`

 So this wrapper will be superfluous, right? :)

JuliaSoboleva:
@wbaumann Will, your commit didn't break anything, all looks good (especially I noted nice updated edit text fields).

Thus, to finish Drag&Drop task at first I need to change merge logic and then implement UI changes for Columns fragment and for Receipts.

app/src/main/java/co/smartreceipts/android/persistence/database/tables/ReceiptsTable.java
wbaumann: @JuliaSoboleva good catch. It is! Thanks :)

wbaumann:
@JuliaSoboleva Sounds good overall. I'll merge this code in with the rest of the app once you've added the database merge logic in :). We can do the rest with separate PRs. Thanks!

JuliaSoboleva:
@wbaumann Hi Will,
I've updated merging process. Now backup database is upgraded to current database version at first. 

Also, I've completely changed `Categories`, `PaymentMethods`, `CSV` and `PDF` tables merging logic. I decided to use `ATTACH DATABASE` query and now we can really merge old and new data, not just drop current tables and create new ones. And code for these operations began to look more compact :)

app/src/main/java/co/smartreceipts/android/persistence/DatabaseHelper.java
```diff
@@ -476,9 +496,166 @@ public final synchronized boolean merge(String dbPath, String packageName, boole
...
+                // Merge PDF
+                Logger.debug(this, "Merging PDF");
+
+                final String mergePDF = "INSERT INTO " + PDFTable.TABLE_NAME +
+                        "( " + PDFTable.COLUMN_TYPE + ", " + AbstractSqlTable.COLUMN_LAST_LOCAL_MODIFICATION_TIME + ", " + AbstractSqlTable.COLUMN_CUSTOM_ORDER_ID + ")" +
+                        " SELECT backup_pdf." + PDFTable.COLUMN_TYPE + ", \"" + System.currentTimeMillis() + "\"" + ", \"" + Integer.MAX_VALUE + "\"" +
+                        " FROM backup_db." + PDFTable.TABLE_NAME + " backup_pdf " +
+                        " LEFT JOIN " + PDFTable.TABLE_NAME + " current_pdf " +
+                        " ON backup_pdf." + PDFTable.COLUMN_TYPE + " = current_pdf." + PDFTable.COLUMN_TYPE +
+                        " WHERE current_pdf." + PDFTable.COLUMN_TYPE + " IS NULL" +
```
wbaumann: Just curious - what's the purpose of the `WHERE` clause here for the `COLUMN_TYPE IS NULL` check?

JuliaSoboleva: Good question :)
I've found this kind of hack [here](http://zametkinapolyah.ru/wp-content/uploads/2016/07/type-join.png).  

In the table that `LEFT JOIN` returns, there will be all rows from the `left` table to which the corresponding values from the `right` table will be added even if these values are `NULL`. This step allows us to select such rows after `LEFT JOIN` where `left` table items have no correspondences in the `right` table. So when the `left` table is `backup_db.PDFTable` and the `right` is `current_db.PDFTabe`, we can choose such `COLUMN_TYPE`s from backup table that are not present in the current table.  

Now I think that I could use `WHERE` with `COLUMN_ID IS NULL`, it might look better. But actually, this has no sense because if no match is found, then all values will be `NULL`.

wbaumann: @JuliaSoboleva Makes sense! Great find and thank you for the update

PR #62, opened 2017-11-23
Drag and drop for CSV and PDF columns

JuliaSoboleva:
Hi Will, 

Happy Thanksgiving day! :)

I've finished with drag and drop for CSV and PDF columns fragment. This looks very similar to `Categories` and `Payment methods` reordering that we made before.

The next and apparently the last which we need to make draggable are `Receipts`. And I have a couple of questions about it:

- The first problem is `ListFragment`. `ReceiptsListFragment` is based on `ListFragment` and we can migrate it to `RecycleView` fragment (`RecyclerView` is required for drag and drop). But there are also `DistanceFragment` and `TripFragment` that are based on `ListFragment`. Should we migrate just `ReceiptsListFragment` or better to make it for all?

- The second question is how reordering of receipts will look like. I assume that adding a `Reorder` option to the main menu (which contains currently `Settings`, `OCR` and `Backups`) is quite not obvious to the user. I think the best I can propose for now is to add permanent `drag handle view` for `Receipt` item and perform dragging on long click on this handle (not on the whole item like it's implemented for categories, payment methods, and columns now). What do you think about it? Maybe you have better ideas? :)


wbaumann:
@JuliaSoboleva Thank you for the Thanksgiving wishes and great work! :)

For your questions:

* Good point. Let's start by making the `Receipts` use a `RecyclerView` for sorting. We'll probably eventually want to support this behavior for `DistanceFragment` and `TripFragment` as well, but no users have really requested that, so I'm fine with holding off on those in the short term (though feel free to migrate them to using a `RecyclerView` as well). And of course if it makes sense to change everything at once, go for it!
* As for the drag handle on long press, I think that makes sense. If it would make it easier, one of my goals is to change the receipt UI to modernize it at some point with more of a card style layout. I think it'd be useful to remove the majority of item in the receipt menu (eg "Edit", "Move/Copy", etc) in favor of supplying these controls in a more direct way. For example, we could pin the receipt image  image [similar to this example](https://storage.googleapis.com/material-design/publish/material_v_12/assets/0B6Okdz75tqQsM3ZqMmJXdjRyZk0/components-cards9.png) and launch the `ReceiptImageFragment` whenever the user clicks on the image (or if no image, launch a file picker). Similarly, tapping on the item itself could reveal the `CreateEditReceiptFragment` directly. There are some good screenshots here that I've been using for inspiration: https://www.getapp.com/blog/best-android-finance-apps/. Overall, I'm more than happy to alter the underlying card UI of the `Receipt` as you see fit, so feel free to move things around if you think it would help with exposing the re-ordering controls. But overall, I think your idea makes sense :)

Thanks!

app/src/main/java/co/smartreceipts/android/persistence/database/tables/ordering/OrderingPreferencesManager.java
```diff
@@ -55,6 +77,14 @@ public boolean isPaymentMethodsTableOrdered() {
         return getSharedPreferences().getBoolean(Keys.ORDERING_PAYMENT_METHODS, false);
     }
 
+    public boolean isCsvColumnsOrdered() {
```
wbaumann: Question on these:

* What happens if the user imports a backup that has column ordering but then does not open the `Manage PDF/CSV Columns" screen? Will it use the custom ordering? If not, we may want to add a control to the `AppVersionManager` such that any time a user performs a new install (or similiar) we automatically update these to set the value to `true`. What do you think?

JuliaSoboleva: Hey Will, sorry for a delay. It's not the easiest question :)

Currently, after merging process unique columns from backup will be added to existing columns (to default columns in our case because user doesn't open "Manage PDF/CSV Columns" fragment). And it will use default ordering. So, as I understand, the order is not guaranteed but most likely items that were inserted later will be located at the end. If backup contains the same default columns set but reordered, then user loses custom ordering.

If we will just add that control, I think we won't get needed result because default `custom_order_id = 0` and unique items which will be merged from backup will have `custom_order_id = Integer.MAX_VALUE` so `orderBy custom_order_id` is also can't guarantee order.

Now it comes to my mind that currently `overwrite` option for column tables is not implemented. Actually, I think the best we can do to restore ordering clearly is to drop current columns table and save imported table (if user checked `overwrite` checkbox) and then call `OrderingPreferencesManager#savePdf(Csv)ColumnsTableOrdering`.

 - Thus, without `overwrite` checkbox: we are getting columns from backup that are not present in default columns set and adding them to default columns (order is default, new columns are at the end);

- With `overwrite` checkbox: current table is replaced with the table from backup (with all order numbers), and we save flag that says to use `orderBy custom_order_id`.

Do you agree?

wbaumann: @JuliaSoboleva That approach makes sense. If I recall, I `overwrite` for the column and category tables was never implemented, since it didn't make sense to do so (in favor of always overwriting them). 

My concern with the first approach is that you could import an old backup without `overwrite` and then have your actual columns added to the end of the default list (that you get on first install). My loose preference would just be to apply the latter approach:

> With `overwrite` checkbox: current table is replaced with the table from backup (with all order numbers), and we save flag that says to use `orderBy` `custom_order_id`.

For both `overwrite` and no `overwrite` cases. I think that's safer for the columns. More generally, I'm planning to remove the `overwrite` behavior from the import, since I've found it to confuse users (eg some things it deletes all data, some think it changes only parts). 

Does that work? Thanks!

JuliaSoboleva: Thanks for clarifying!

Yes, I agree with you. Don't you mind if I'll push this changes to the next PR with receipt's drag&drop?

wbaumann: @JuliaSoboleva No problem at all. Thanks Julia!

PR #65, opened 2018-01-30
Bugfixes and new dialog for file attachment

JuliaSoboleva:
I fixed a few bugs related to date picker behaviour (same-day reports and strange receipts ordering).

Also, I added a new dialog for adding an attachment (on empty file tap) which shows three options: 
* "Take a photo"
* "Pick a photo"
* "Attach file"

And one new option for removing the attachment.

By the way, I added to `dimens` some unified values for margins, paddings and font sizes. I hope that this change will simplify our work a bit :)


app/src/main/java/co/smartreceipts/android/di/SmartReceiptsActivityBindingModule.java
```diff
@@ -87,7 +87,7 @@
 
     @Binds
     @IntoMap
-    @FragmentKey(CreateEditReceiptFragment.class)
+    @FragmentKey(ReceiptCreateEditFragment.class)
```
wbaumann: Not related to this, but we can upgrade our Dagger to 2.11+ to take advantage of the new `@ContributesAndroidInjector` annotation, which removes the need for the `@Binds`, `@IntoMap`, `@FragmentKey` boilerplate. 

https://proandroiddev.com/dagger-2-annotations-binds-contributesandroidinjector-a09e6a57758f

app/src/main/java/co/smartreceipts/android/receipts/ReceiptsListFragment.java
```diff
@@ -412,10 +426,43 @@ protected PersistenceManager getPersistenceManager() {
         return persistenceManager;
     }
 
+    private void showAttachmentDialog(final Receipt receipt) {
```
wbaumann: Is it possible to make this launch a `DialogFragment` instead? The code looks good (and we may be limited with the `highlightedReceipt` stuff), but that'll insure that this can handle rotation. Does the following flow work otherwise:

1. Launch `showAttachmentDialog`
2. Wait for the dialog to appear
3. Rotate

I think the dialog will be auto-dismissed in this case? Switching to a `DialogFragment` will handle this for us.

app/src/main/java/co/smartreceipts/android/receipts/ReceiptsListFragment.java
```diff
@@ -442,25 +489,42 @@ public final boolean showReceiptMenu(final Receipt receipt) {
...
                         final DeleteReceiptDialogFragment deleteReceiptDialogFragment = DeleteReceiptDialogFragment.newInstance(receipt);
                         navigationHandler.showDialog(deleteReceiptDialogFragment);
+
                     } else if (selection.equals(receiptActionMoveCopy)) {// Move-Copy
                         analytics.record(Events.Receipts.ReceiptMenuMoveCopy);
                         ReceiptMoveCopyDialogFragment.newInstance(receipt).show(getFragmentManager(), ReceiptMoveCopyDialogFragment.TAG);
+
+                    } else if (selection.equals(receiptActionRemoveAttachment)) { // Remove Attachment
+                        analytics.record(Events.Receipts.ReceiptMenuRemoveAttachment);
+                        new AlertDialog.Builder(getActivity())
```
wbaumann: Same - can we make this a `DialogFragment` to better handle rotations?

app/src/main/java/co/smartreceipts/android/receipts/attacher/ReceiptAttachmentManager.java
```diff
@@ -0,0 +1,58 @@
...
+        analytics.record(Events.Receipts.ReceiptAttachPhoto);
+        return new CameraInteractionController(fragment).addPhoto();
+    }
+
+    public boolean attachPicture(Fragment fragment) {
+        analytics.record(Events.Receipts.ReceiptAttachPicture);
+
+        try {
+            final Intent intent = new Intent(Intent.ACTION_GET_CONTENT);
+            intent.addCategory(Intent.CATEGORY_OPENABLE);
```
wbaumann: Not sure if it makes sense, but could we have the `ImportPhotoPdfDialogFragment` class also call this?

app/src/main/java/co/smartreceipts/android/receipts/editor/ReceiptCreateEditFragment.java
```diff
@@ -186,10 +188,10 @@
     @BindView(R.id.receipt_input_tax_wrapper)
     View taxInputWrapper;
 
-    @BindViews({R.id.receipt_input_guide_image_payment_method, R.id.receipt_input_payment_method })
+    @BindViews({R.id.receipt_input_guide_image_payment_method, R.id.receipt_input_payment_method})
     List<View> paymentMethodsViewsList;
 
-    @BindViews({R.id.receipt_input_guide_image_exchange_rate, R.id.receipt_input_exchange_rate, R.id.receipt_input_exchanged_result, R.id.receipt_input_exchange_rate_base_currency })
+    @BindViews({R.id.receipt_input_guide_image_exchange_rate, R.id.receipt_input_exchange_rate, R.id.receipt_input_exchanged_result, R.id.receipt_input_exchange_rate_base_currency})
```
wbaumann: Once we move to Kotlin, all this comes for free :)

app/src/main/res/drawable/ic_attach_file_white_24dp.xml
```diff
@@ -0,0 +1,9 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
wbaumann: Thanks for this!

app/src/main/res/layout/backups_header.xml
```diff
@@ -4,149 +4,144 @@
...
         android:layout_height="wrap_content"
-        android:layout_marginBottom="@dimen/default_margin_item_separator"
+        android:layout_marginBottom="@dimen/margin_tiny"
         android:text="@string/manual_backup_description" />
 
-    <RelativeLayout
+    <LinearLayout
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
+        android:layout_marginBottom="@dimen/margin_tiny"
```
wbaumann: And all this!

app/src/main/java/co/smartreceipts/android/di/GlobalBindingModule.java
```diff
@@ -1,193 +1,128 @@
...
 public abstract class GlobalBindingModule {
-    @Binds
-    @IntoMap
-    @ActivityKey(SmartReceiptsActivity.class)
-    public abstract AndroidInjector.Factory<? extends Activity> smartReceiptsActivitySubcomponentBuilder(
-            SmartReceiptsActivitySubcomponent.Builder builder);
-
-    @Binds
-    @IntoMap
-    @ActivityKey(SettingsActivity.class)
```
wbaumann: Nice :)

app/src/main/java/co/smartreceipts/android/fragments/ImportPhotoPdfDialogFragment.java
```diff
@@ -34,21 +45,32 @@ public Dialog onCreateDialog(Bundle savedInstanceState) {
...
+        } else if (which == WHICH_PDF) {
+            attachmentResult = receiptAttachmentManager.attachFile(getParentFragment(), true);
+        }
+
+        if (!attachmentResult) {
             Toast.makeText(getContext(), getString(R.string.error_no_file_intent_dialog_title), Toast.LENGTH_SHORT).show();
         }
+
+//
+//        try {
```
wbaumann: Can we delete this code rather than commenting it out? We can always use `git blame` to figure out what happened

app/src/main/java/co/smartreceipts/android/receipts/ReceiptsListFragment.java
```diff
@@ -412,10 +427,21 @@ protected PersistenceManager getPersistenceManager() {
         return persistenceManager;
     }
 
+    private void showAttachmentDialog(final Receipt receipt) {
+        highlightedReceipt = receipt;
+
+        final ReceiptAttachmentDialogFragment attachmentDialogFragment = ReceiptAttachmentDialogFragment.newInstance(receipt);
+        attachmentDialogFragment.setTargetFragment(this, 0);
+
+        attachmentDialogFragment.show(getFragmentManager(), ReceiptAttachmentDialogFragment.class.getSimpleName());
```
wbaumann: I think we can use `getChildFragmentManager`, so we can avoid the above to call to `setTargetFragment` above.

app/src/main/java/co/smartreceipts/android/receipts/ReceiptsListFragment.java
```diff
@@ -672,6 +705,11 @@ public void displayOcrStatus(@NonNull UiIndicator<String> ocrStatusIndicator) {
         }
     }
 
+    @Override
+    public void setImageUri(Uri uri) {
```
wbaumann: `@NonNull`? Or `@Nullable`?

app/src/main/java/co/smartreceipts/android/receipts/attacher/ReceiptAttachmentDialogFragment.java
```diff
@@ -0,0 +1,98 @@
...
+        super.onCreate(savedInstanceState);
+        receipt = getArguments().getParcelable(Receipt.PARCEL_KEY);
+        Preconditions.checkNotNull(receipt, "ReceiptAttachmentDialogFragment requires a valid Receipt");
+    }
+
+    @NonNull
+    @Override
+    public Dialog onCreateDialog(Bundle savedInstanceState) {
+
+        final Fragment targetFragment = getTargetFragment();
```
wbaumann: Here's what we can potentially change if we can use the `getParentFragment`. No worries if not :)

app/src/main/java/co/smartreceipts/android/receipts/attacher/ReceiptAttachmentDialogFragment.java
```diff
@@ -0,0 +1,98 @@
...
+                .setNegativeButton(android.R.string.cancel, (dialog, which) -> dialog.cancel());
+
+        final View dialogView = getActivity().getLayoutInflater().inflate(R.layout.dialog_receipt_attachment, null);
+        dialogBuilder.setView(dialogView);
+
+        final AlertDialog dialog = dialogBuilder.create();
+
+        dialogView.findViewById(R.id.attach_photo).setOnClickListener(v -> {
+            ((Listener) targetFragment).setImageUri(receiptAttachmentManager.attachPhoto(getTargetFragment()));
+            dialog.cancel();
```
wbaumann: Let's use the `dismiss()` call instead:

https://developer.android.com/guide/topics/ui/dialogs.html#DismissingADialog

app/src/main/java/co/smartreceipts/android/receipts/attacher/ReceiptAttachmentDialogFragment.java
```diff
@@ -0,0 +1,98 @@
...
+        });
+
+        dialogView.findViewById(R.id.attach_picture).setOnClickListener(v -> {
+            if (!receiptAttachmentManager.attachPicture(getTargetFragment(), false)) {
+                Toast.makeText(getContext(), getString(R.string.error_no_file_intent_dialog_title), Toast.LENGTH_SHORT).show();
+            }
+            dialog.cancel();
```
wbaumann: Same for here

app/src/main/java/co/smartreceipts/android/receipts/attacher/ReceiptAttachmentDialogFragment.java
```diff
@@ -0,0 +1,98 @@
...
+        });
+
+        dialogView.findViewById(R.id.attach_file).setOnClickListener(v -> {
+            if (!receiptAttachmentManager.attachFile(getTargetFragment(), false)) {
+                Toast.makeText(getContext(), getString(R.string.error_no_file_intent_dialog_title), Toast.LENGTH_SHORT).show();
+            }
+            dialog.cancel();
```
wbaumann: And here

app/src/main/java/co/smartreceipts/android/receipts/attacher/ReceiptAttachmentManager.java
```diff
@@ -0,0 +1,60 @@
...
+public class ReceiptAttachmentManager {
+
+    private final Analytics analytics;
+
+    @Inject
+    ReceiptAttachmentManager(Analytics analytics) {
+        this.analytics = analytics;
+    }
+
+    public Uri attachPhoto(Fragment fragment) {
```
wbaumann: `@NonNull` return type

app/src/main/java/co/smartreceipts/android/receipts/attacher/ReceiptAttachmentManager.java
```diff
@@ -0,0 +1,60 @@
...
+        analytics.record(Events.Receipts.ReceiptAttachPhoto);
+        return new CameraInteractionController(fragment).addPhoto();
+    }
+
+    public boolean attachPicture(Fragment fragment, boolean newReceipt) {
```
wbaumann: `@NonNull Fragment fragment`

app/src/main/java/co/smartreceipts/android/receipts/attacher/ReceiptRemoveAttachmentDialogFragment.java
```diff
@@ -0,0 +1,71 @@
...
+
+
+public class ReceiptRemoveAttachmentDialogFragment extends DialogFragment {
+
+    @Inject
+    ReceiptTableController receiptTableController;
+
+    private Receipt receipt;
+
+    // TODO: 05.02.2018 remove attachment file manually. All files are saved to app's dir
```
wbaumann: Can we delete these?

app/src/main/java/co/smartreceipts/android/receipts/attacher/ReceiptRemoveAttachmentDialogFragment.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    @NonNull
+    @Override
+    public Dialog onCreateDialog(Bundle savedInstanceState) {
+        return new AlertDialog.Builder(getActivity())
+                .setTitle(receipt.getName())
+                .setCancelable(true)
+                .setMessage(getString(R.string.receipt_dialog_remove_attachment))
+                .setNegativeButton(android.R.string.cancel, (dialogRemove, id) -> dialogRemove.cancel())
```
wbaumann: `dismiss()` here too

JuliaSoboleva:
@wbaumann Thanks for your review and all your comments! :)
I fixed all you asked for, but that TODO regarding removing files was actual and I also handled it already. 

But I also probably found some bug, and I want to deal with it before pushing updates to you.
The thing is while retaking receipt's photo we are saving new image with a name `System.currentTimeMillis() + ".jpg"` then making some duplicate and it seems like we're forgetting about that file. Here's an [example](https://photos.app.goo.gl/QS7jssvXMXNiQfrD3) - duplicated file and garbage after files that were already deleted before. Is it right that this behaviour isn't expected and we need to delete the duplicated file?

And one more question. Now we have "Take Receipt Image" option in Receipt's menu but I think it's needless now since we have `ReceiptAttachmentDialogFragment`. Can we delete this option from the menu?

wbaumann:
@JuliaSoboleva Good idea - let's delete "Take Receipt Image", since we have the `ReceiptAttachmentDialogFragment`. You're right that this will solve the same case :)

And good catch! Thank you. Yes - let's delete that extra file. Here's the reason behind it:

* Users wanted their file names to show the receipt name itself (eg `#_Name.jpg`)
* But when the user first takes an image, he/she has not entered the name yet
* To account for this, we save the image with a "random" name (timestamp) and then rename it after the fact

I think we can fix this by directly overwriting the receipt name when retaking the image, since the user cannot change the name or number at that point. Would that work for the fix? And thank you for finding this!

JuliaSoboleva:
@wbaumann  Hi Will,

Sorry for this delay. I found out that we've got this issue with garbage files when I worked on permission request at `ReceiptsListFragment`. I was forced to move `activityFileResultLocator.getUriStream()` subscription to `onCreate`. So while `ReceiptImageFragment` was doing his job, `ReceiptsListFragment` was also listening to coming `Uri`'s. And because of this, we have had 2 images that were saved to app's folder from two different threads and one of them were ignored next.

I solved it in not such pretty way, but I didn't find another one. I removed code block which saves the image from `ReceiptImageFragment` with an eye to the fact that this task will be carried out in `ReceiptsListFragment`.

I also found some problem with Picasso cashing, but it's already done.

Thank you for your patience

wbaumann:
@JuliaSoboleva No worries about the delay. Thank you so much for working on this and taking the time to find and fix the issues. Reviewing now :)

wbaumann:
@JuliaSoboleva 

Awesome work! This all looks really great, and I’m going to merge this in as it stands. Thank you so much for taking the time to get this all done properly :)

With that said, I also want to think over the case in which we’re now commenting out the call to:

```
activityFileResultImporter.importFile(locatorResponse.getRequestCode(),
                                locatorResponse.getResultCode(), locatorResponse.getUri(), receipt.getTrip());
```

Your approach will definitely work, but I’m concerned about what happens if the `ReceiptsListFragment` gets evicted from memory for whatever reason. Walking through the code, it seems that we need to handle the following cases:

1. When a user imports “file://“ content, we may need to initiate a permissions request. Since the permissions prompt moves the fragment into an `onPause` (or `onStop`?) state, we cannot drive the Rx stream via an `onResume` subscription (as we’ll miss the `checkPermissionAndMaybeAsk` result).
2. To account for this, we register the `ActivityFileResultLocator` in the `onCreate` method of both the `ReceiptsListFragment` and `ReceiptImageFragment`. Both receive the same event, however, which can cause a “duplicate” file to be saved to our file-system. You fixed this issue in the current CR by commenting out that line for `ReceiptImageFragment`.

What do you think about the following:

* Moving everything back into `onResume`

This **should** work (untested of course), because the call to `permissionsDelegate.checkPermissionAndMaybeAsk(READ_PERMISSION)` will always check `permissionStatusChecker.isPermissionGranted(manifestPermission)` first. So even if we miss the "granted" event, we'll immediately get the valid result as soon as we re-subscribe. Or am I missing something here?

I’ll move this to Trello for easier tracking / discussion. Thanks again!

PR #67, opened 2018-03-01
Several bug fixes

JuliaSoboleva:
* Handled import file flow. I changed `PermissionRequesterHeadlessFragment` a little bit, so now we need to call `markRequestConsumed(@NonNull String manifestPermission)` method to mark permission request result as not important. Thus now all works good and I returned import implementation to `ReceiptImageFragment` as it should be.

* Changed receipt layout (no more markers and new position for category).

* Added date subheaders for receipts list.

* FIxed bug with "next" key while using the keyboard.  Simple adding `android:imeOptions="actionNext"` to `EditText`es helped :)

* Added current trip highlighting for tablets.

* Fixed a few bugs for tablets:
    - Bug with unnormal trip card height. It occurred because of dynamic price width calculation. Calculation is still dynamic but now it works well on both phones and tablets :)
    - Bug with keeping several `ReportInfoFragment` instances in back stack. When you clicked at Trip1, Trip2, Trip3 - 3 `ReportInfoFragment`s were saved to back stack and you needed to tap on back button 3 times to close all `ReportInfoFragment`s.

Thanks in advance for your review and have a great day :)


app/src/main/java/co/smartreceipts/android/activities/NavigationHandler.java
```diff
@@ -63,6 +64,12 @@ public void navigateUpToTripsFragment() {
 
     public void navigateToReportInfoFragment(@NonNull Trip trip) {
         if (isDualPane) {
+            // we don't need to keep several ReportInfoFragment instances in backstack
```
wbaumann: Great find :). Thank you for this!

app/src/main/java/co/smartreceipts/android/imports/intents/widget/info/IntentImportInformationInteractor.java
```diff
@@ -51,18 +52,19 @@ public IntentImportInformationInteractor(@NonNull IntentImportProcessor intentIm
...
                                             return Maybe.just(intentImportResult);
                                         } else {
-                                            return permissionRequester.request(Manifest.permission.READ_EXTERNAL_STORAGE)
+                                            return permissionRequester.request(READ_PERMISSION)
                                                     .flatMapMaybe(permissionAuthorizationResponse -> {
                                                         if (permissionAuthorizationResponse.wasGranted()) {
                                                             return Maybe.just(intentImportResultReference);
                                                         } else {
-                                                            return Maybe.error(new PermissionsNotGrantedException("User failed to grant READ permission", Manifest.permission.READ_EXTERNAL_STORAGE));
+                                                            permissionRequester.markRequestConsumed(READ_PERMISSION);
```
wbaumann: How come we have it here too? I think it's fine to just have it in the `subscribe` part right? Or no?

app/src/main/java/co/smartreceipts/android/receipts/ReceiptsListFragment.java
```diff
@@ -257,7 +253,45 @@ public void onResume() {
...
+                        return Single.just(response);
+                    } else { // we need to check read external storage permission
+                        Logger.debug(this, "need to check permission");
+                        return permissionsDelegate.checkPermissionAndMaybeAsk(READ_PERMISSION)
+                                .toSingleDefault(response)
+                                .onErrorReturn(ActivityFileResultLocatorResponse::LocatorError);
+                    }
+                })
+                .subscribe(locatorResponse -> {
+                    permissionsDelegate.markRequestConsumed(READ_PERMISSION);
```
wbaumann: Meaning this subscribe one

app/src/main/java/co/smartreceipts/android/imports/intents/widget/info/IntentImportInformationInteractor.java
JuliaSoboleva: This `markRequestConsumed` is from another use case. We don't use `PermissionsDelegate` in `IntentImportInformationInteractor` and `IntentImportInformationPresenter` so we need to mark permission request here

wbaumann: Ah makes sense. Thanks!

PR #71, opened 2018-03-26
Kotlin support and receipt image sharing

JuliaSoboleva:
There are:
* small bugfix for `MatchReceiptNameToCategory` and `MatchReceiptCommentToCategory`
* small fix for `distance_no_data` string resources
* new "Share" option for `ReceiptImageFragment`
* Kotlin support (I also converted a few files to Kotlin)

I wish `ImmutableCategoryImpl` was a data class but it has a bit custom `equals` implementation (`syncState` is ignored).

One more thing about Kotlin. I added Kotlin Source Set and now we have `src/main/kotlin`. The project is large enough and navigation between `.kt` and `.java` files seemed to me not very comfortable. 
Anyway, take a look. Interesting to know what do you think about it :)


build.gradle
```diff
@@ -1,4 +1,5 @@
 buildscript {
+    ext.kotlin_version = '1.2.31'
```
wbaumann: Let's do this in another pull request, but we can move this value to block at line 27 below:

```
ext {
     minSdkVersion = 21
     targetSdkVersion = 26
     compileSdkVersion = 26
     buildToolsVersion = "26.0.2"
}
```

JuliaSoboleva: I tried to move `kotlin_version = '1.2.31'` to `ext` but for some reason, it doesn't work. Probably `ext` global object is not available inside `buildscript` block

PR #72, opened 2018-04-12
Fixes for reports and updated reordering logic for receipts

JuliaSoboleva:
Fixed couple bugs from reports generation: 
- Now total `Price` column contains a list of prices with different currencies
- `Currency` total column also contains a list of currency codes for multiple currencies
- Fixed total `Price(exchanged)` value rounding

Also, changed the logic of formation `customOrderId` for Receipts.
Now `customOrderId = receipt.getDate().toDays() + receipt_number_among_same_days` (more detailed scheme [here](https://trello-attachments.s3.amazonaws.com/53eac8f36e04c457940d90b2/5ac40f53bddc96017bdebd43/7358e59ac66362626273d52d506d9a94/Screenshot.png))

After this update users who already used receipts reordering can face a problem - all their `customOrderId`s will be discarded to new default values and receipts will be sorted by date (I added toast for this situation). But after this change they will be able to use reordering normally.


app/src/main/java/co/smartreceipts/android/date/DateUtils.java
```diff
@@ -70,4 +70,8 @@ public static boolean isToday(@NonNull Date date) {
 
         return startTime <= testTime && testTime <= endTime;
     }
+
+    public static long getDays(@NonNull Date date) {
+        return date.getTime() / (1000 * 60 * 60 * 24);
```
wbaumann: We can replace this with:

`return date.getTime() / TimeUnit.DAYS.toMillis(1);`

app/src/main/java/co/smartreceipts/android/model/impl/ImmutableNetPriceImpl.java
```diff
@@ -149,14 +158,27 @@ public String getCurrencyCodeFormattedPrice() {
...
-            final List<String> currencyStrings = new ArrayList<>();
-            for (PriceCurrency currency : currencyToPriceMap.keySet()) {
-                currencyStrings.add(ModelUtils.getCurrencyCodeFormattedValue(currencyToPriceMap.get(currency), currency));
-            }
-            return TextUtils.join("; ", currencyStrings);
+            return getCurrencyCodeFormattedStringFromMap(currencyToPriceMap);
+        }
+    }
+
+    public String getCurrencyCodeFormattedNotExchangedPrice() {
```
wbaumann: Should we replace the `getCurrencyCodeFormattedPrice` with this option implementation?

app/src/main/java/co/smartreceipts/android/model/impl/columns/receipts/ReceiptPriceColumn.java
```diff
@@ -39,7 +40,14 @@ public String getFooter(@NonNull List<Receipt> receipts) {
             for (final Receipt receipt : receipts) {
                 prices.add(receipt.getPrice());
             }
-            return new PriceBuilderFactory().setPrices(prices, tripCurrency).build().getDecimalFormattedPrice();
+
+            final Price netPrice = new PriceBuilderFactory().setPrices(prices, tripCurrency).build();
+            if (netPrice instanceof ImmutableNetPriceImpl) {
+                return ((ImmutableNetPriceImpl) netPrice).getCurrencyCodeFormattedNotExchangedPrice();
```
wbaumann: Meaning for this one - we could just return the `getCurrencyCodeFormattedPrice` value?

app/src/main/java/co/smartreceipts/android/model/impl/columns/receipts/ReceiptTaxColumn.java
```diff
@@ -35,7 +36,14 @@ public String getFooter(@NonNull List<Receipt> receipts) {
             for (final Receipt receipt : receipts) {
                 prices.add(receipt.getTax());
             }
-            return new PriceBuilderFactory().setPrices(prices, tripCurrency).build().getDecimalFormattedPrice();
+
+            final Price netPrice = new PriceBuilderFactory().setPrices(prices, tripCurrency).build();
+            if (netPrice instanceof ImmutableNetPriceImpl) {
+                return ((ImmutableNetPriceImpl) netPrice).getCurrencyCodeFormattedNotExchangedPrice();
```
wbaumann: Same here

app/src/main/java/co/smartreceipts/android/receipts/ReceiptsAdapter.java
```diff
@@ -101,18 +104,39 @@ public long getItemId(int position) {
...
+
+        Logger.debug(this, "Moving receipt from position {} to position {}", realFromPosition, realToPosition);
+
+        if (!orderingPreferencesManager.isReceiptsTableOrdered()) setDefaultOrderIds();
+
+        if (isOldCustomOrderIdFormatUsed()) {
+            setDefaultOrderIds();
+            writeUpdatedItems(items);
+
+            Toast.makeText(context, "Sorry, your receipts now are ordered by date.", Toast.LENGTH_SHORT).show();
```
wbaumann: We can probably just remove this. I think it'll be okay if we just make this change once without the error message

app/src/main/java/co/smartreceipts/android/receipts/ReceiptsAdapter.java
```diff
@@ -101,18 +104,39 @@ public long getItemId(int position) {
...
+            return;
         }
+
+        List<Receipt> originalItemsList = new ArrayList<>(items);
+
+        final Receipt movedReceipt = items.remove(realFromPosition);
+        items.add(realToPosition, movedReceipt);
+        updateListItems();
+
+        final List<Receipt> changedReceipts = ReceiptCustomOrderIdHelper.INSTANCE
```
wbaumann: Can we use Dagger to inject `ReceiptCustomOrderIdHelper` instead of the `ReceiptCustomOrderIdHelper.INSTANCE`

app/src/main/java/co/smartreceipts/android/settings/widget/editors/DraggableEditableListFragment.java
```diff
@@ -81,6 +82,14 @@ public boolean onOptionsItemSelected(MenuItem item) {
         return true;
     }
 
+    /**
+     * super.saveTableOrdering must be called
```
wbaumann: Let's annotate this with the `CallSuper` instead

app/src/main/kotlin/co/smartreceipts/android/receipts/helper/ReceiptCustomOrderIdHelper.kt
```diff
@@ -0,0 +1,51 @@
+package co.smartreceipts.android.receipts.helper
+
+import co.smartreceipts.android.model.Receipt
+import co.smartreceipts.android.model.factory.ReceiptBuilderFactory
+
+object ReceiptCustomOrderIdHelper {
```
wbaumann: Kotlin - Nice :)

app/src/test/kotlin/co/smartreceipts/android/model/impl/ImmutableNetPriceImplTest.kt
```diff
@@ -0,0 +1,231 @@
...
+            String.format("%s; %s", EUR_CURRENCY.currencyCode, USD_CURRENCY.currencyCode),
+            differentCurrenciesWithExchangeRatePrice.currencyCode
+        )
+    }
+
+    @Test
+    fun getCurrencyCodeFormattedNotExchangedPriceTest() {
+        assertEquals("$3.00", sameCurrencyPrice.currencyCodeFormattedNotExchangedPrice)
+        assertEquals(
+            "EUR1.00; USD2.00",
```
wbaumann: And thank you for adding tests to these items!

app/src/main/java/co/smartreceipts/android/settings/widget/editors/DraggableEditableListFragment.java
```diff
@@ -81,6 +82,14 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
+     */
+    protected void saveTableOrdering() {
+        Logger.debug(this, "saveTableOrdering");
+        ((DraggableEditableCardsAdapter)adapter).saveNewOrder(getTableController());
```
wbaumann: Rather than casting, can we ensure that the `adapter` is a `DraggableEditableCardsAdapter` when defined?

app/src/main/kotlin/co/smartreceipts/android/receipts/helper/ReceiptCustomOrderIdHelper.kt
wbaumann: I also uploaded your diagram to S3 here:

https://s3.amazonaws.com/smartreceipts/Diagrams/SmartReceiptsCustomSortingOrderDesign.png

So we can include this in the KDocs

app/src/main/java/co/smartreceipts/android/receipts/ReceiptsAdapter.java
wbaumann: The main benefit to avoid `INSTANCE` is for unit tests. It's not immediately useful, but it eventually makes these harder, since there's not a simple way to ever inject a `mock(INSTANCE)` for testing purposes. With that said, this is a relatively minor item so not a blocker. Thank you for all the fixes :)

PR #79, opened 2018-06-03
Hebrew translations

tomgs:
Hiya,

Will - as per our e-mail conversation, first part of the Hebrew translation's here.

Will work on it bit by bit until it's done:)

Should I translate the two preferences files as well?

PR #80, opened 2018-07-12
Report preferred language

JuliaSoboleva:
Added new option that allows user to choose a language for generated reports. According to this next changes took place:
* Added new setting for Preferred Report Language in the Report section. Default value is current app language. 
* Changed column tables structure (`PdfTable` and `CsvTable`). In previous version, it kept column header as `String`, now we can use int for column type. Added migration implementation which allows saving previous user settings (except cases when current phone language is not the same as was when the column table was created)
* A bit more files are converted to Kotlin :)

By the way, unfortunately, I didn't manage `ReportResourcesManager` unit testing because of this Roboelectric [issue](https://github.com/robolectric/robolectric/issues/2909)


wbaumann:
Thank you @JuliaSoboleva . I'll review shortly. Also - please ignore the Travis CI build failure. I started adding this but didn't complete the setup in a functional manner, so it's nothing on your end

app/src/main/java/co/smartreceipts/android/persistence/database/tables/AbstractColumnTable.java
```diff
@@ -1,62 +1,62 @@
...
-    private final String mIdColumnName;
-    private final String mTypeColumnName;
+    public static final String COLUMN_ID = "id";
+    public static final String COLUMN_TYPE = "column_type";
+
+    private final int tableExistsSinceDatabaseVersion;
+    private final ColumnDefinitions<Receipt> receiptColumnDefinitions;
+
+    @Deprecated
+    public static final String idColumnName = "id";
```
wbaumann: Let's make these all capital letters like the other statics: `DEPRECATED_COLUMN_ID_AS_NAME`

app/src/main/java/co/smartreceipts/android/persistence/database/tables/AbstractColumnTable.java
```diff
@@ -1,62 +1,62 @@
...
+    @Deprecated
+    public static final String typeColumnName = "type";
```
wbaumann: Same

app/src/main/java/co/smartreceipts/android/persistence/database/tables/AbstractColumnTable.java
```diff
@@ -91,28 +91,84 @@ public synchronized void onUpgrade(@NonNull SQLiteDatabase db, int oldVersion, i
...
+                    final int idIndex = columnsCursor.getColumnIndex(idColumnName);
+                    final int typeIndex = columnsCursor.getColumnIndex(typeColumnName);
+
+                    do {
+                        final int id = columnsCursor.getInt(idIndex);
+                        final String oldColumnType = columnsCursor.getString(typeIndex);
+
+                        int newColumnType = receiptColumnDefinitions.getDefaultInsertColumn().getType();
+
+                        if (receiptColumnDefinitions instanceof ColumnFinder) {
```
wbaumann: Rather than doing this `instanceOf` check, can we just pass the `ColumnFinder` to the constructor? Since we know our `receiptColumnDefinitions` always implement `ColumnFinder`?

Then we can do something like:

```
private final receiptColumnsDefinitions;
private final columnFinder;

ColumnsTable(receiptColumnsDefinitions) {
    this(receiptColumnsDefinitions, (ColumnFinder) receiptColumnsDefinitions);
}

ColumnsTable(receiptColumnsDefinitions, columnFinder) {
    this.receiptColumnsDefinitions = notNull(receiptColumnsDefinitions);
    this.columnFinder = notNull(columnFinder);
}
```

app/src/main/java/co/smartreceipts/android/persistence/database/tables/AbstractColumnTable.java
```diff
@@ -91,28 +91,84 @@ public synchronized void onUpgrade(@NonNull SQLiteDatabase db, int oldVersion, i
...
+                                newColumnType = columnTypeByHeaderValue;
+                            }
+                        }
+
+                        final ContentValues columnValues = new ContentValues(1);
+                        columnValues.put(COLUMN_TYPE, newColumnType);
+                        Logger.debug(this, "Updating old column header value: {} to new column type {}", oldColumnType, newColumnType);
+
+                        if (db.update(getTableName(), columnValues, COLUMN_ID + "= ?", new String[]{Integer.toString(id)}) == 0) {
+                            Logger.error(this, "Column update error happened");
```
wbaumann: Let's throw an exception here, since I'd prefer that we crash the app (to properly resolve this) rather than silently masking the error and potentially causing the user to lose his/her columns

app/src/main/java/co/smartreceipts/android/settings/UserPreferenceManager.java
```diff
@@ -96,6 +96,15 @@ public void initialize() {
                                 context.getResources().getValue(userPreference.getDefaultValue(), typedValue, true);
                                 preferences.get().edit().putFloat(preferenceName, typedValue.getFloat()).apply();
                                 Logger.debug(UserPreferenceManager.this, "Assigned default float value for {} as {}", preferenceName, typedValue.getFloat());
+                            } else if (UserPreference.ReportOutput.PreferredReportLanguage.equals(userPreference)) {
+                                final Locale currentLocale = context.getResources().getConfiguration().locale;
```
wbaumann: I think we can just use `Locale.getDefault()` here instead

app/src/main/java/co/smartreceipts/android/settings/UserPreferenceManager.java
```diff
@@ -96,6 +96,15 @@ public void initialize() {
...
+                                final String[] supportedLanguages = context.getResources().getStringArray(R.array.pref_output_preferred_language_entryValues);
+                                for (String supportedLanguage : supportedLanguages) {
+                                    if (currentLocale.getLanguage().equals(supportedLanguage)) {
+                                        preferences.get().edit().putString(preferenceName, currentLocale.getLanguage()).apply();
```
wbaumann: What happens if we never set this? Will it default to English?

<Nevermind this: Answered my own question below. It will >

app/src/main/kotlin/co/smartreceipts/android/model/impl/columns/AbstractColumnImpl.kt
```diff
@@ -0,0 +1,38 @@
...
+
+import co.smartreceipts.android.model.ActualColumnDefinition
+import co.smartreceipts.android.model.Column
+import co.smartreceipts.android.sync.model.SyncState
+import java.util.*
+
+/**
+ * Provides an abstract implementation of the column contract to cover shared code
+ */
+abstract class AbstractColumnImpl<T> @JvmOverloads constructor(
```
wbaumann: Kotlin makes these classes much easier :)

app/src/main/kotlin/co/smartreceipts/android/model/impl/columns/distance/DistancePriceColumn.kt
```diff
@@ -0,0 +1,28 @@
...
+) : AbstractColumnImpl<Distance>(id, DistanceColumnDefinitions.ActualDefinition.PRICE, syncState) {
+
+    override fun getValue(distance: Distance): String? =
+        if (allowSpecialCharacters) distance.price.currencyFormattedPrice
+        else distance.price.currencyCodeFormattedPrice
+
+    override fun getFooter(distances: List<Distance>): String {
+        val tripCurrency = if (!distances.isEmpty()) distances[0].trip.tripCurrency else null
+        return if (allowSpecialCharacters) {
+            PriceBuilderFactory().setPriceables(distances, tripCurrency!!).build()
```
wbaumann: To make things are bit cleaner, want to remove the `80` characters per line limit for this project? Then we won't have to add these new lines for the extra bit. Thanks :)

app/src/main/kotlin/co/smartreceipts/android/model/impl/columns/distance/DistancePriceColumn.kt
```diff
@@ -0,0 +1,28 @@
...
+import co.smartreceipts.android.sync.model.SyncState
+
+class DistancePriceColumn(
+    id: Int,
+    syncState: SyncState,
+    private val allowSpecialCharacters: Boolean
+) : AbstractColumnImpl<Distance>(id, DistanceColumnDefinitions.ActualDefinition.PRICE, syncState) {
+
+    override fun getValue(distance: Distance): String? =
+        if (allowSpecialCharacters) distance.price.currencyFormattedPrice
```
wbaumann: No concern here but in the future, let's always use braces for if/else statements:

```
if (allowSpecialCharacters) {
  distance.price.currencyFormattedPrice
} else {
  distance.price.currencyCodeFormattedPrice
}
```

It can be an occasional source of bugs if we ever chance this to a multi-line condition in the future

app/src/main/kotlin/co/smartreceipts/android/model/impl/columns/receipts/ReceiptColumnDefinitions.kt
```diff
@@ -0,0 +1,212 @@
...
+        TAX_EXCHANGED(19, R.string.column_item_converted_tax_exchange_rate, R.string.original_column_item_converted_tax_exchange_rate_en_us_name),
+        PRICE_PLUS_TAX_EXCHANGED(20, R.string.column_item_converted_price_plus_tax_exchange_rate, R.string.original_column_item_converted_price_plus_tax_exchange_rate_en_us_name),
+        PRICE_MINUS_TAX_EXCHANGED(21, R.string.column_item_converted_price_minus_tax_exchange_rate),
+        EXCHANGE_RATE(22, R.string.column_item_exchange_rate, R.string.original_column_item_exchange_rate_en_us_name),
+        PICTURED(23, R.string.column_item_pictured, R.string.original_column_item_pictured_en_us_name),
+        REIMBURSABLE(24, R.string.column_item_reimbursable, R.string.original_column_item_reimbursable_en_us_name, R.string.column_item_deprecated_expensable),
+        INDEX(25, R.string.column_item_index, R.string.original_column_item_index_en_us_name),
+        ID(26, R.string.column_item_id, R.string.original_column_item_id_en_us_name),
+        PAYMENT_METHOD(27, R.string.column_item_payment_method),
+        EXTRA_EDITTEXT_1(28, R.string.RECEIPTMENU_FIELD_EXTRA_EDITTEXT_1),
```
wbaumann: Let's make these like `100`, `101`, and `102`, since they're all specialty fields for some custom builds. That way we can always keep them at the end

app/src/main/kotlin/co/smartreceipts/android/workers/reports/ReportResourcesManager.kt
```diff
@@ -0,0 +1,54 @@
...
+    context: Context,
+    private val preferenceManager: UserPreferenceManager,
+    private val flex: Flex
+) {
+
+    private var localizedContext: Context
+
+    init {
+        val desiredLocale =
+            Locale(preferenceManager.get(UserPreference.ReportOutput.PreferredReportLanguage))
```
wbaumann: I'm worried this block will run before `userPreferenceManager.initialize()` gets called (which runs in the background and can change the desired locale). Maybe we can just move this whole block to when the user calls `getLocalizedContext`? Thoughts?

app/src/test/kotlin/co/smartreceipts/android/workers/reports/csv/CsvTableGeneratorTest.kt
```diff
@@ -25,14 +22,23 @@ class CsvTableGeneratorTest {
...
 
+    @Mock
+    lateinit var reportResourceManager: ReportResourcesManager
+
     @Before
     @Throws(Exception::class)
     fun setUp() {
         MockitoAnnotations.initMocks(this)
-        `when`(column.header).thenReturn(HEADER)
+        `when`(column.headerStringResId).thenReturn(HEADER_ID)
```
wbaumann: I added a special kotlin-mockito test framework, so you can user `whenever(...)` instead of `when` to avoid the need for the special syntax with \`

app/src/main/res/values/not_translatable.xml
```diff
@@ -32,4 +32,13 @@
...
 
+    <!--Supported languages-->
+    <string name="language_en_entry" translatable="false">en</string>
+    <string name="language_de_entry" translatable="false">de</string>
+    <string name="language_es_entry" translatable="false">es</string>
+    <string name="language_fr_entry" translatable="false">fr</string>
+    <string name="language_pt_entry" translatable="false">pt</string>
+    <string name="language_ru_entry" translatable="false">ru</string>
+    <string name="language_uk_entry" translatable="false">uk</string>
+
```
wbaumann: Let's also add a Hebrew entry (ie `language_iw_entry`). Another contributor has been helping to supply these translations for us:

https://github.com/wbaumann/SmartReceiptsLibrary/pull/79

app/src/main/kotlin/co/smartreceipts/android/workers/reports/ReportResourcesManager.kt
JuliaSoboleva: I agree. Fixed this part

PR #81, opened 2018-07-21
Finished Hebrew Translations

tomgs:
Hey @wbaumann,

All done with the Hebrew translations!

A few things I didn't translate:

1. The `preference_defaults.xml` file, since it seems to contain only boolean fields. 
2. Any fields with `translatable="false"` in them.
3. Any mention of the app's name - it doesn't translate well to Hebrew, in my opinion. קבלות חכמות is the literal translation (Ka-ba-lot Kha-kha-mot, phonetically - ["kha" is the very Hebrew-ey "heit" sound](https://en.wikipedia.org/wiki/Heth#Hebrew_%E1%B8%A4et)), which sounds a bit like a marketing scheme by a not-so-clever advertising agency. I think it's best we leave it at that until we have a better name.
4. The `pref_about_about_dialogMessage` field - due to #3 above.
5. The `pref_about_terms_dialogMessage` field (the GNU Affero license) - I think translating messes up the subtle legalese involved, so I left it at that for now. There's probably a good translation somewhere - I just can't seem to find it! 

Cheers,

Tom

### DIFF
--- a/app/src/main/res/values-he/preference_defaults.xml
+++ b/app/src/main/res/values-he/preference_defaults.xml
@@ -1,0 +1,69 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <!-- ============= Preference General ================= -->
+    <integer name="pref_general_trip_duration_defaultValue">3</integer>
+    <string name="pref_general_default_currency_defaultValue" />
+    <string name="pref_general_default_date_separator_defaultValue" />
+    <bool name="pref_general_track_cost_center_defaultValue">false</bool>
+
+    <!-- ============== Preference Receipt ================ -->
+    <item name="pref_receipt_minimum_receipts_price_defaultValue" format="float" type="dimen">-1000.0</item>
+    <bool name="pref_receipt_predict_categories_defaultValue">true</bool>
+    <bool name="pref_receipt_match_name_to_category_defaultValue">false</bool>
+    <bool name="pref_receipt_match_comment_to_category_defaultValue">false</bool>
+    <bool name="pref_receipt_reimbursable_only_defaultValue">false</bool>
+    <bool name="pref_receipt_reimbursable_default_defaultValue">true</bool>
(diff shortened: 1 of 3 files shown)


tomgs:
Travis fails for some reason... that's a knows issue?

app/src/main/res/values-he/preference_defaults.xml
```diff
@@ -0,0 +1,69 @@
...
+
+    <!-- ============= Preference Distance ================= -->
+    <item name="pref_distance_rate_defaultValue" format="float" type="dimen">0.0</item>
+    <bool name="pref_distance_include_price_in_report_defaultValue">false</bool>
+    <bool name="pref_distance_print_table_defaultValue">true</bool>
+    <bool name="pref_distance_print_daily_defaultValue">false</bool>
+    <bool name="pref_distance_as_price_defaultValue">true</bool>
+
+    <!-- ================= Preference Pro ================== -->
+    <string name="pref_pro_pdf_footer_defaultValue">Report Generated using Smart Receipts for Android</string>
```
wbaumann: This one. But it's okay to stay in English too :)

wbaumann:
And just for reference, I made two minor commits to fix up a few things:

* https://github.com/wbaumann/SmartReceiptsLibrary/commit/9ae7df3db6245c0748781603b00301fe886e91f2
* https://github.com/wbaumann/SmartReceiptsLibrary/commit/7ff2ce3ea5cf0a8946bb9d47d06c28118b2d6399

The second one does the real work (eg replacing `?` with `/?` in Android studio to address some build warnings). Great work though! Thank you!

tomgs:
Sure man, no worries. Was fun - got to see how the translations play out on Android. Quite a nice way of keeping the translations organised and only show what's needed.

wbaumann:
Agreed. Android's management system is pretty solid (or at least easier than iOS in my opinion), but it's good exposure. Looking over your email next :)

PR #83, opened 2018-09-24
Add UUID column for all tables

JuliaSoboleva:
It took longer than I expected, but it's finally done!

Thank you for your instructions about secret files! :)

As about changes from my side:

* A bunch of models is converted to Kotlin, I also changed it to use `@Parcelize` and removed `Parcel` boilerplate code.
* I removed some interfaces (like `Category`, `Receipt`, etc.)
* Updated DB version to 19. Added `entity_uuid` column to all tables. To make this change more global I also decided to change Trips table primary key from `TEXT` to `INTEGER AUTOINCREMENT` (like all other tables have). Trip name (which is used as a directory name) is still unique. 
* Random UUID is assigned to entity during `AbdstractSqlTable#incertBlocking`. I also tried assigning random UUID while creating an object at `BuilderFactory` but it turns testing to a very difficult part so I decided to keep this logic on a table level. So now `BuilderFactory` uses default UUID value `UUID.fromString("00000000-0000-0000-0000-000000000000")`. Also, during `AbdstractSqlTable#updateBlocking` we're verifying that UUID value isn't changed.
* `id` and `uuid` properties now belong to `Syncable` interface. I thought it's a rational change but maybe `Syncable` name isn't perfect for this, I'm not sure.

And I left a couple of new TODOs:
* I didn't implement updates for merge process, I'm not sure how should it look like.
* Now all our tables have integer autoincrement primary key but I'm not sure if I can remove all code for different primary key types. Maybe it'll be needed in the future, I don't know. What do you think about it?

Thanks!


wbaumann:
@JuliaSoboleva Thanks for the great work for this! I'll start my review shortly but for your general notes:

- Nice work! And thanks for taking advantage of `Parcelize` to make our lives easier :)
- Thanks for removing these interfaces!
- Perfect. And makes sense about the `TripsTable` change as well
- And great - I figured that's the safest source, so we're safe in case someone ever forgets by accident! As for the builder factory, is there a reason to use `UUID.fromString("00000000-0000-0000-0000-000000000000")` vs `UUID.random()` (or whatever the new random one is)?
- Could we create a new interface instead of `Syncable` for this? I'd rather have `Syncable` be specific for the Google Drive (and future providers) code. Maybe something like `Identifiable` or `Keyed` or for these two methods?

As for the TODOs:

- For the merge process, this is for the database right? I can tackle this after you've merged it in, so no worries. My thinking is that we'd use the `entity_uuid` when merging, taking the item that's more recent. 
- Yea I think we can remove our primary key code. Eventually I want to migrate over to using [Room](https://developer.android.com/topic/libraries/architecture/room), but I'll wait for things to settle before doing so. Plus we can always revert those changes if we need in the future :)... But let's do that in a future PR, so we can keep this one smaller for review.

Thank you as always!

wbaumann:
Also - looks like Travis is failing due to a merge conflict between our code:

https://travis-ci.com/wbaumann/SmartReceiptsLibrary/builds/85707897

I had changed `ReceiptColumnDefinitions.getColumnFromDefinition` to be public to simplify some things for the white-labelled versions of the app. You may need to make that public again to fix the other builds. Thanks!

app/src/main/java/co/smartreceipts/android/currency/PriceCurrency.java
```diff
@@ -114,4 +120,28 @@ public int hashCode() {
...
+    @Override
+    public void writeToParcel(Parcel dest, int flags) {
+        dest.writeString(getCurrencyCode());
+
+    }
+
+    public static final Creator<PriceCurrency> CREATOR = new Creator<PriceCurrency>() {
+        @Override
+        public PriceCurrency createFromParcel(Parcel in) {
+            return new PriceCurrency(in);
```
wbaumann: Could we change this to call:

```
PriceCurrency.getInstance(in.readString())
```

Instead of:

```
new PriceCurrency(in);
```

The former will take advantage of some caching that we use, since we create a lot of these instances :)

app/src/main/java/co/smartreceipts/android/activities/FragmentProvider.java
```diff
@@ -50,7 +50,7 @@ public TripFragment newTripFragmentInstance() {
     @NonNull
     public ReportInfoFragment newReportInfoFragment(@NonNull Trip trip) {
         Bundle args = new Bundle();
-        args.putParcelable(Trip.PARCEL_KEY, trip);
+        args.putParcelable(Trip.Companion.getPARCEL_KEY(), trip);
```
wbaumann: For each of these, we can update the `Trip.PARCEL_KEY` field to add the `@JvmStatic` annotation, and then we can continue to reference it as before (i.e. `Trip.PARCEL_KEY` instead of `Trip.Companion.getPARCEL_KEY()`).

This walks through this some more: https://kotlinlang.org/docs/reference/java-to-kotlin-interop.html

app/src/main/java/co/smartreceipts/android/model/factory/ColumnBuilderFactory.java
```diff
@@ -17,13 +20,15 @@
...
+    private UUID uuid;
     private int columnType;
     private SyncState syncState;
     private long customOrderId;
 
     public ColumnBuilderFactory(@NonNull ColumnDefinitions<T> columnDefinitions) {
         this.columnDefinitions = columnDefinitions;
-        id = Column.UNKNOWN_ID;
+        id = Syncable.MISSING_ID;
+        uuid = Syncable.Companion.getMISSING_UUID();
```
wbaumann: Same here - we can use the `@JvmStatic` annotation (in `Syncable` or whatever) to make this a little cleaner for the Java side

app/src/main/java/co/smartreceipts/android/persistence/database/restore/ByRowDatabaseMerger.java
```diff
@@ -112,6 +113,7 @@ public Completable merge(@NonNull DatabaseHelper currentDatabase, @NonNull Datab
             final List<Trip> importedTrips = new ArrayList<>(importedBackupDatabase.getTripsTable().getBlocking());
             Logger.info(ByRowDatabaseMerger.this, "Importing {} trip entries", importedTrips.size());
             for (final Trip importedTrip : importedTrips) {
+                // TODO: 31.08.2018 trip names are still unique (because of directory names). what if imported trip has same name but different UUID?
```
wbaumann: Good question... In this case, let's just merge the data from both trips into a single trip with the same name. We can then do the UUID for each of the distances and receipts. This seems safest, so we accidentally don't delete one of someone's reports

app/src/main/java/co/smartreceipts/android/persistence/database/tables/AbstractSqlTable.java
```diff
@@ -116,6 +119,40 @@ protected synchronized void onUpgradeToAddSyncInformation(@NonNull SQLiteDatabas
...
+                    do {
+                        final int id = cursor.getInt(idIdx);
+                        final String uuid = UUID.randomUUID().toString();
+
+                        final ContentValues columnValues = new ContentValues(1);
+                        columnValues.put(COLUMN_UUID, uuid);
+                        Logger.debug(this, "Updating UUID value to {}", uuid);
+
+                        if (db.update(getTableName(), columnValues, COLUMN_ID + "= ?", new String[]{Integer.toString(id)}) == 0) {
+                            Logger.error(this, "Column update error happened");
```
wbaumann: Let's make this throw an `IllegalStateException` if this fails. I'm sure we'll revert to your logging fix soon, but this will help me in the short-term to identify any potential bugs that we introduce

app/src/main/java/co/smartreceipts/android/persistence/database/tables/AbstractSqlTable.java
```diff
@@ -36,6 +37,8 @@
  */
 public abstract class AbstractSqlTable<ModelType, PrimaryKeyType> implements Table<ModelType, PrimaryKeyType> {
```
wbaumann: Could you update this (or maybe the generic `Table` interface to use:

`Table<ModelType extends OurNewIdentifiableInterface>`?

Then we can assume that we always have access to `getId()` and `getUuid()`

app/src/main/java/co/smartreceipts/android/settings/widget/editors/categories/CategoriesListFragment.java
```diff
@@ -116,11 +116,12 @@ private void showCreateEditDialog(@Nullable Category editCategory) {
...
                     final String newCode = codeBox.getText().toString();
 
-                    final Category category = new CategoryBuilderFactory()
+                    final CategoryBuilderFactory categoryBuilderFactory = new CategoryBuilderFactory()
                             .setName(newName)
                             .setCode(newCode)
-                            .setCustomOrderId(isEdit ? editCategory.getCustomOrderId() : Long.MAX_VALUE)
-                            .build();
+                            .setCustomOrderId(isEdit ? editCategory.getCustomOrderId() : Long.MAX_VALUE);
+                    if (isEdit) categoryBuilderFactory.setUuid(editCategory.getUuid());
```
wbaumann: Can we wrap this if statement in braces like:

```
if (isEdit) {
  // ...
}
```

It just makes things a bit safer for us

app/src/main/java/co/smartreceipts/android/sync/drive/managers/DriveRestoreDataManager.java
```diff
@@ -225,37 +226,52 @@ public DriveRestoreDataManager(@NonNull Context context,
...
-                if (cursor != null && cursor.moveToFirst()) {
-                    final int driveIdIndex = cursor.getColumnIndex(AbstractSqlTable.COLUMN_DRIVE_SYNC_ID);
-                    final int parentIndex = cursor.getColumnIndex(ReceiptsTable.COLUMN_PARENT);
-                    final int pathIndex = cursor.getColumnIndex(ReceiptsTable.COLUMN_PATH);
+                final String[] selection = new String[]{AbstractSqlTable.COLUMN_DRIVE_SYNC_ID, ReceiptsTable.COLUMN_PARENT_TRIP_ID, ReceiptsTable.COLUMN_PATH};
+                receiptsCursor = importDb.query(ReceiptsTable.TABLE_NAME, selection, AbstractSqlTable.COLUMN_DRIVE_SYNC_ID + " IS NOT NULL AND " + ReceiptsTable.COLUMN_PATH + " IS NOT NULL AND " + AbstractSqlTable.COLUMN_DRIVE_MARKED_FOR_DELETION + " = ?", new String[]{Integer.toString(0)}, null, null, null);
+                if (receiptsCursor != null && receiptsCursor.moveToFirst()) {
+                    final int driveIdIndex = receiptsCursor.getColumnIndex(AbstractSqlTable.COLUMN_DRIVE_SYNC_ID);
+                    final int parentIdIndex = receiptsCursor.getColumnIndex(ReceiptsTable.COLUMN_PARENT_TRIP_ID);
```
wbaumann: This is actually going to be tricky. There's no guarantee that our old backups will have this column changes in their table. Is it possible to still have it query off the parent trip name as a fallback? Something like:

```
if (hasParentTripId) {
  // New approach
} else {
  // Old approach
}
```

app/src/main/kotlin/co/smartreceipts/android/model/Distance.kt
```diff
@@ -0,0 +1,135 @@
...
+     */
+    val timeZone: TimeZone,
+    /**
+     * The user defined comment [String] for this receipt
+     */
+    val comment: String
+) : Parcelable, Priceable, Comparable<Distance>, Syncable {
+
+    companion object {
+        val PARCEL_KEY: String = Distance::class.java.name
```
wbaumann: For this - you can use the `@JvmStatic` to have it reference in the old way

app/src/test/java/co/smartreceipts/android/persistence/database/tables/ReceiptsTableTest.java
```diff
@@ -21,6 +22,7 @@
 import java.util.Collections;
 import java.util.List;
 import java.util.TimeZone;
+import java.util.UUID;
 
 import co.smartreceipts.android.model.Category;
 import co.smartreceipts.android.model.PaymentMethod;
```
wbaumann: For this class, the `DistanceTableTest` and the `TripsTableTest`, could you add an `onUpgradetoV19` type test that runs through the core logic for these? If you're having trouble just let me know, and I'll do it. I'm just always a bit more careful with any fundamental db changes :)

app/src/main/kotlin/co/smartreceipts/android/sync/model/Syncable.kt
```diff
@@ -1,12 +1,34 @@
...
  * Marks a particular model object as capable of being synced with a remote server environment
  */
 interface Syncable {
 
+    /**
+     * The primary key id for this item
+     *
+     * @return the items's autoincrement id
+     */
+    val id: Int
```
wbaumann: As noted in the comments, let's move this and the `uuid` into a different interface

app/src/main/kotlin/co/smartreceipts/android/sync/model/Syncable.kt
```diff
@@ -1,12 +1,34 @@
...
+     */
+    val uuid: UUID
+
     /**
      * The current [SyncState] associated with this item
      */
     val syncState: SyncState
+
+    companion object {
+        const val MISSING_ID: Int = -1
```
wbaumann: And we can use the `@JvmStatic` for both of these

app/src/main/java/co/smartreceipts/android/currency/PriceCurrency.java
wbaumann: Thanks:)

app/src/main/java/co/smartreceipts/android/activities/FragmentProvider.java
wbaumann: Ah good to know - I didn't know that either, so good to learn!

app/src/main/java/co/smartreceipts/android/sync/drive/managers/DriveRestoreDataManager.java
wbaumann: Thank you 👍

wbaumann:
Makes sense about all the tests for `BuilderFactory.build()`. I'll look to see if there's an easier/better way to manage this!

app/src/main/java/co/smartreceipts/android/persistence/database/tables/AbstractSqlTable.java
```diff
@@ -290,28 +336,45 @@ public synchronized final void onPostCreateUpgrade() {
...
+        if (!(oldModelType instanceof Syncable)) {
+            return Optional.absent();
+        }
+
         final ContentValues values = databaseAdapter.write(newModelType, databaseOperationMetadata);
-        final String oldPrimaryKeyValue = primaryKey.getPrimaryKeyValue(oldModelType).toString();
+
+        // to be sure that entity_uuid will never be changed
+        if (values.containsKey(COLUMN_UUID)) {
+            values.remove(COLUMN_UUID);
```
wbaumann: Could you update this to also add a `Logger.warn` statement if we ever do this? I want to work through updating these flows. Thanks!

app/src/main/java/co/smartreceipts/android/persistence/database/tables/AbstractSqlTable.java
```diff
@@ -243,6 +281,13 @@ public synchronized final void onPostCreateUpgrade() {
     @NonNull
     public synchronized Optional<ModelType> insertBlocking(@NonNull ModelType modelType, @NonNull DatabaseOperationMetadata databaseOperationMetadata) {
         final ContentValues values = databaseAdapter.write(modelType, databaseOperationMetadata);
+
+        // to be sure that entity_uuid is not missed
+        if (!values.containsKey(COLUMN_UUID) || UUID.fromString(values.getAsString(COLUMN_UUID)).equals(Keyed.Companion.getMISSING_UUID())) {
+            values.put(COLUMN_UUID, UUID.randomUUID().toString());
```
wbaumann: Could you update this to also add a `Logger.warn` statement if we ever do this? I want to work through updating these flows. Thanks!

app/src/main/java/co/smartreceipts/android/persistence/database/tables/AbstractSqlTable.java
```diff
@@ -274,7 +319,8 @@ public synchronized final void onPostCreateUpgrade() {
                 }
             } else {
                 // If it's not an auto-increment id, just grab whatever the definition is...
-                final ModelType insertedItem = databaseAdapter.build(modelType, primaryKey, databaseOperationMetadata);
+                // TODO: 17.09.2018 all tables have autoincrement key id now...
```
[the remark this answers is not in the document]
JuliaSoboleva: I'll remove this TODO now and make refactoring with next PR :)

app/src/main/java/co/smartreceipts/android/persistence/database/tables/AbstractSqlTable.java
```diff
@@ -290,28 +336,45 @@ public synchronized final void onPostCreateUpgrade() {
 
     @SuppressWarnings("unchecked")
     public synchronized Optional<ModelType> updateBlocking(@NonNull ModelType oldModelType, @NonNull ModelType newModelType, @NonNull DatabaseOperationMetadata databaseOperationMetadata) {
+
+        if (!(oldModelType instanceof Syncable)) {
```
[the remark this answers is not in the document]
JuliaSoboleva: All our models are `Syncable` so it'll never fall actually, but it's my bad I'll return old code.

app/src/main/java/co/smartreceipts/android/persistence/database/tables/AbstractSqlTable.java
JuliaSoboleva: Added `Logger.warn`

app/src/main/java/co/smartreceipts/android/persistence/database/tables/AbstractSqlTable.java
JuliaSoboleva: Added `Logger.warn`

app/src/main/java/co/smartreceipts/android/persistence/database/tables/AbstractSqlTable.java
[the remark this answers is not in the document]
wbaumann: Thanks 👍 . One thing we can also do it ensure that the parent table has <ModelType extends Keyed & Syncable> to ensure that we handle this without all the `instanceof` stuff. I can take care of this in a side PR too.

app/src/main/java/co/smartreceipts/android/persistence/database/tables/AbstractSqlTable.java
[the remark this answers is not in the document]
wbaumann: Cool. Thank you and great work with this!

PR #85, opened 2018-10-03
Added support for gradle parallelization to help improve unit test pe…

wbaumann:
…rformance


wbaumann:
Some notes: 

1. Running with 4 threads takes > 19 minutes
2. Running with 2 threads takes ~16 minutes
3. Running with 1 thread takes ~15-16 minutes

As a result, I'm going to abandon this PR. It seems to be related to the same issue detailed here:

https://github.com/robolectric/robolectric/issues/3351

PR #115, opened 2018-10-18
final working version of translation string conversion tool + travis.…

tomgs:
…yml changes for the dependencies.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,6 +43,10 @@ notifications:
     on_failure: always # default: always
 
 before_install:
+- sudo add-apt-repository -y ppa:cpick/hub
+- sudo apt-get -y update
+- sudo apt-get -y install hub
+- gem install twine
 - mkdir "$ANDROID_HOME/licenses" || true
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -43,6 +43,10 @@ notifications:
     on_failure: always # default: always
 
 before_install:
+- sudo add-apt-repository -y ppa:cpick/hub
```
wbaumann: Could you add a comment for each of these as to what it's used for? Just for quick reference

scripts/barebones-locale-tool.sh
```diff
@@ -134,20 +134,24 @@ rm en-diff.txt
 #printf "Removed us.lproj.\n"
 
 ## Push changes to repo
-printf "Start Interacting With GitHub\n"
+printf "\nStart Interacting With GitHub\n"
 echo "-----------------------"
 cd SmartReceiptsIOS
+echo "Getting acess token via cURL"
```
wbaumann: Want to also add a comment here like:

`# We're ignoring the blatant security hole here, since this service account does not have permissions to merge in this code without review anyway`

PR #134, opened 2019-02-04
Organization support [Part 2]

JuliaSoboleva:
Now we have full implementation of checking and applying all `Organization`'s settings (it was supported for app's preferences before, now added for `Categories`, `Payment Methods`, csv and pdf `Columns`).

Also, I removed a mocked value for `column_type` according to the server fix.
BTW, we still have mocked user role in the `AccountFragment`.

And one more thing. You've updated `StrictModeConfiguration` and added custom `penaltyListener` with
```
if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
    ... addind listener
} else {
    threadPolicyBuilder.penaltyDeath()
}
```
I'm using Android 8.1 now and app crushes while inflating, so I changed `penaltyDeath()` to `penaltyLog()` :)


app/src/main/java/co/smartreceipts/android/identity/organization/AppPreferencesSynchronizer.kt
```diff
@@ -0,0 +1,112 @@
...
+            .map { someSettingsDontMach -> !someSettingsDontMach }
+    }
+
+    internal fun applyOrganizationPreferences(organizationSettings: AppSettings.OrganizationPreferences): Completable {
+        return Single.just(organizationSettings)
+            .flatMapCompletable { settings ->
+                userPreferenceManager.userPreferencesSingle
+                    .flatMapObservable<UserPreference<*>> { Observable.fromIterable(it) }
+                    .map { userPreference -> applyPreference(settings, userPreference) }
+                    .ignoreElements()
```
wbaumann: Is `ignoreElements` just to complete it?

app/src/main/java/co/smartreceipts/android/identity/organization/AppPreferencesSynchronizer.kt
```diff
@@ -0,0 +1,112 @@
...
+        return Single.just(organizationPreferences)
+            .flatMap { settings ->
+                userPreferenceManager.userPreferencesSingle
+                    .flatMapObservable<UserPreference<*>> { Observable.fromIterable(it) }
+                    .map { userPreference -> checkPreferenceMatch(settings, userPreference) }
+                    .filter { it.isPresent }
+                    .map<Boolean> { it.get() }
+                    .contains(false)
+            }
+            .map { someSettingsDontMach -> !someSettingsDontMach }
```
wbaumann: Let's fix this type: `someSettingsDontMach` => `someSettingsDontMatch` with an extra `"t"`

app/src/main/java/co/smartreceipts/android/identity/organization/AppSettingsSynchronizer.kt
```diff
@@ -0,0 +1,152 @@
...
+        return tableController.get()
+            .flatMap { appItems ->
+                Logger.debug(this, "Comparing app settings with organization...")
+                for (organizationItem in organizationItems) {
+                    var foundSame = false
+
+                    for (appItem in appItems) {
+                        if (equals(appItem, organizationItem)) {
+                            foundSame = true
+                            Logger.debug(this, "Found same item: $appItem")
```
wbaumann: Let's use the SLF4J syntax like `Logger.debug(this, "Found same item: {}", appItem)` instead, since it'll avoid the string interpolation action if we're not actually using this log type

wbaumann:
Thanks @JuliaSoboleva !

PR #136, opened 2019-03-02
Adding Android CircleCI Integration

tomgs:


.circleci/config.yml
```diff
@@ -0,0 +1,74 @@
...
+             printf "\n"
+             echo "-- Pushing To Github --"
+             printf "\n"
+             git config credential.helper 'cache --timeout=120'
+             git config user.email ${GITHUB_EMAIL}
+             git config user.name ${GITHUB_USERNAME}
+             git add ../app/build.gradle
+             git commit -m "Bumping Project Version [ci skip]"
+             cd ..
+             git push https://${GITHUB_API_TOKEN}@github.com/tomgs/SmartReceiptsLibrary.git ${CIRCLE_BRANCH}
```
wbaumann: Let's change this to point to my branch instead :)

.circleci/config.yml
```diff
@@ -0,0 +1,74 @@
+version: 2
```
wbaumann: Why version 2?

.circleci/config.yml
```diff
@@ -0,0 +1,74 @@
...
+    environment:
+      JVM_OPTS: -Xmx3200m
+    steps:
+      - run:
+          name: Wait 900 Seconds (15 Minutes - To Wait For Additional Commits)
+          command: sleep 900
+      - checkout
+      - restore_cache:
+          key: jars-{{ checksum "build.gradle" }}-{{ checksum  "app/build.gradle" }}
+#      - run:
```
wbaumann: Does this fail any more?

.circleci/config.yml
```diff
@@ -0,0 +1,74 @@
...
+      - run:
+          name: Download Dependencies
+          command: ./gradlew androidDependencies
+      - save_cache:
+          paths:
+            - ~/.gradle
+          key: jars-{{ checksum "build.gradle" }}-{{ checksum  "app/build.gradle" }}
+      - run:
+          name: Assemble Bundles & Run Tests
+          command: ./gradlew clean bundleFreeRelease bundlePlusFlavorRelease
```
wbaumann: This definitely runs the tests, correct? I wasn't 100% sure but wanted to check. We can always change this to:

`./gradlew clean test bundleFreeRelease bundlePlusFlavorRelease`

If not

app/build.gradle
```diff
@@ -241,4 +256,11 @@ task translateToIOS << {
      }
  }
 
+play {
+    serviceAccountCredentials = file("GPLAY_KEY.json")
+    track = "beta"
+    userFraction = 0.05
```
wbaumann: Let's remove this for now. I think beta can go to everyone 100% :)

.circleci/config.yml
tomgs: fixed in first commit

.circleci/config.yml
tomgs: As mentioned in our Slack conversation, this is just the current default version. All of the guides and the initial setup is using that version, so I went with it.

.circleci/config.yml
tomgs: I think it's in CircleCI's default config file. Might as well leave it there, just in case.

.circleci/config.yml
tomgs: Interesting question. It runs a lot of stuff, but I'm not a hundred percent sure it runs all the tests. Might as well add the test there, adding now in second commit. If it fails let me know, I haven't run it with my fork.

app/build.gradle
tomgs: Yep, should be a hundred. Changing in third commit.

tomgs:
All done with the changes you requested.

PR #143, opened 2019-04-30
New "Price with Currency Symbol" column

JuliaSoboleva:
* Added a new "Price with Currency Symbol" column to PDF and CSV reports
* Removed currency spinner dropdown chevron for `ReceiptCreateEditFragment` (it caused invisible currency code on small screens)


wbaumann:
Thanks @JuliaSoboleva !

PR #145, opened 2019-06-22
adding check to determine report metadata zip name

HeinousGames:


wbaumann:
Nice fix @HeinousGames :)

PR #147, opened 2019-06-27
Espresso test fixes

HeinousGames:
getDecimalFormattedExchangeRate("USD") expects two extra zeroes. not sure if this is normal behavior or not.


wbaumann:
@HeinousGames Yeah that's fine. I had made that change and then reverted it, so NBD there

PR #148, opened 2019-06-28
Organizations support [Part 3]

JuliaSoboleva:
* added `secrets_extractor.bat` script for Windows systems (from Yooz branch)
* implemented uploading organization settings to the server
* updated `AccountFragment` to present organizations list instead of single "primary" organization, added custom `DiffUtils` implementation for a nice animation when the state of organization changes 

By the way, what do you think, should we implement changing the organization's name from the app (if the user is an admin)?


wbaumann:
@JuliaSoboleva Does the endpoint support this:

> should we implement changing the organization's name from the app (if the user is an admin)?

It's low priority so not a big concern, but go for it if it's easy :)

app/src/main/java/co/smartreceipts/android/identity/widget/account/AccountPresenter.kt
```diff
@@ -24,43 +26,68 @@ class AccountPresenter @Inject constructor(view: AccountView, interactor: Accoun
         )
 
         compositeDisposable.add(
-            interactor.getOrganization()
-                .subscribe { organizationIndicator ->
-                        view.presentOrganization(organizationIndicator)
+            organizationsSubject
+                .subscribe {
+                    view.presentOrganizations(it)
+                    Logger.debug(this, "Updating organizations list")
```
wbaumann: In the future, let's move log statements into `doOnNext` blocks, since they're more side effects

PR #149, opened 2019-06-29
Robolectric update

HeinousGames:


wbaumann:
Thanks!

One thing to keep in the back of your mind for the future - it'd be great if we could have a nightly build job, which auto-updates our dependencies, since our test scripts should do a good job of informing us if anything will fail

PR #150, opened 2019-07-18
Image cropping feature

JuliaSoboleva:
I've integrated [UCrop](https://github.com/Yalantis/uCrop) library (I think it looks richer than cropper that you suggested in Trello, but it's pretty easy to change if you want :) ).

Now after taking/importing a photo for a new/existing receipt, Crop Activity will be launched. If a user cancels cropping we're showing a toast with a message: "You can disable crop image option in the app settings". 

Also, I've changed `ReceiptImageFragment`. I removed all implementation of image rotation (it's done by UCrop) and added new "Edit" button which launches Crop Activity.

Now we have a new `CheckBoxPreference` setting "Enable Crop Image Screen". Cropping is enabled by default.

BTW, I've already mentioned that `StrictMode` crashes the app with `DiskReadViolation` when the camera app is opening, so I added `StrictModeConfiguration#permitDiskReads` function for wrapping "safe" methods.


wbaumann:
Hey one quick note on this:

> BTW, I've already mentioned that StrictMode crashes the app with DiskReadViolation when the camera app is opening, so I added StrictModeConfiguration#permitDiskReads function for wrapping "safe" methods.

One of the things I'd love to do at some point is to improve the camera launch flow, so the `File#read` process happens in the background. Then we can re-enabled this. It'll help us reduce ANRs over time too. No worries for this change but something to keep in mind for the future. I'm reviewing the rest now!

app/src/main/java/co/smartreceipts/android/model/Receipt.kt
```diff
@@ -108,7 +109,7 @@ class Receipt constructor(
      * was updated between two "like" receipts
      */
     val fileLastModifiedTime: Long
-        get() = file?.lastModified() ?: -1
+        get() = StrictModeConfiguration.permitDiskReads { file?.lastModified() } ?: -1
```
wbaumann: Nice. This works for now :)

wbaumann:
@JuliaSoboleva Nice work as always! I'll move this into Trello but for:

> Now after taking/importing a photo for a new/existing receipt, Crop Activity will be launched. If a user cancels cropping we're showing a toast with a message: "You can disable crop image option in the app settings".

Could we do this as a `Tooltip` in the receipts editor to make it easier? Rules can be if the cropping option has ever been shown previously, we show that with the highest priority. The Copy can be:

"Enable Cropping   Yes | No"

PR #155, opened 2019-08-12
Google drive api refactoring

HeinousGames:
Things to keep in mind regarding individual classes as we move towards using the public drive folder:

- DriveClientInitializer:
    - Wifi Settings for uploads not tracked through drive client anymore. Look into resumable file uploads as an alternative.
- DriveFilesAndFoldersPrinter:
    - logAllFilesAndFolders method wasn’t used before, so I didn’t update it, just commented it out.
- RxDriveTask:
    - Removed methods reliant on drive dependencies
- GoogleDriveBackupManager:
    - Contains async task to check token validity, can it be refactored?
- Drive Service Helper:
    -  Google's custom class (https://github.com/gsuitedevs/android-samples/blob/master/drive/deprecation/app/src/main/java/com/google/android/gms/drive/sample/driveapimigration/DriveServiceHelper.java) was copied in and modified to fit our needs. Some options that are available to us may be better to use than what I am using.
- Five classes in the services module were removed for reliance on deprecated library objects, as well as the test class tied to the functionality of those classes. Do we still need functionality that those classes provided?


wbaumann:
Starting a review and will finish in a couple hours when I'm back. Some initial notes:

> DriveFilesAndFoldersPrinter:
>       logAllFilesAndFolders method wasn’t used before, so I didn’t update it, just commented it out.

It's not "used", but I do occasionally use it whenever I create a custom build for clients that need extra debugging. Could you add the `@SuppressWarning` instead?

> GoogleDriveBackupManager:
>     Contains async task to check token validity, can it be refactored?

I'll look. In general, I try to avoid `AsyncTask`s at all costs, so it's definitely a good place

>    Drive Service Helper:
>        Google's custom class (https://github.com/gsuitedevs/android-samples/blob/master/drive/deprecation/app/src/main/java/com/google/android/gms/drive/sample/driveapimigration/DriveServiceHelper.java) was copied in and modified to fit our needs. Some options that are available to us may be better to use than what I am using.

Sounds good. I'll take a look :)

>    Five classes in the services module were removed for reliance on deprecated library objects, as well as the test class tied to the functionality of those classes. Do we still need functionality that those classes provided?

I'll check these out shortly! Thanks!

app/build.gradle
```diff
@@ -131,7 +135,14 @@ android {
 
         implementation "com.android.support.constraint:constraint-layout:$CONSTRAINT_LAYOUT_VERSION"
 
-        implementation "com.google.android.gms:play-services-drive:$DRIVE_API_VERSION"
+        implementation 'com.google.http-client:google-http-client-gson:1.26.0'
```
wbaumann: Let's move our versioning into the `gradle.properties` file, so we can access them as variables (like `$DRIVE_API_VERSION`

app/src/main/java/co/smartreceipts/android/di/GlobalBindingModule.java
```diff
@@ -44,10 +42,6 @@
     @ContributesAndroidInjector
     public abstract SettingsActivity settingsActivity();
 
-    @ServiceScope
-    @ContributesAndroidInjector
-    public abstract DriveCompletionEventService driveCompletionEventService();
```
wbaumann: This class existed, since Google Drive would "commit" to uploading something, but it wouldn't be official until they sent an event that it succeeded. I added this service to understand when it was actually done, which allowed us to access their ID in a stable (i.e. non-null) manner. Hopefully the new library avoids the need for this :)

app/src/main/java/co/smartreceipts/android/sync/drive/rx/RxDriveTask.java
```diff
@@ -0,0 +1,108 @@
...
+import co.smartreceipts.android.sync.utils.SyncSchedulers;
+import io.reactivex.Completable;
+import io.reactivex.Observable;
+import io.reactivex.Single;
+
+/**
+ * Provides us with some Rx extensions around the {@link Task} object that is used for Google Drive
+ * integrations
+ */
+public class RxDriveTask {
```
wbaumann: If we convert this to Kotlin, we can use extension functions to make this a bit prettier, but not required :)

app/src/main/java/co/smartreceipts/android/sync/drive/GoogleDriveBackupManager.java
```diff
@@ -332,33 +330,21 @@ private void onGoogleSignInAccountReady(@NonNull GoogleSignInAccount signInAccou
...
-                            GoogleSignInAccountFinalization(signInAccount);
-                        }
-                    }
-                };
-
-                task.execute();
+                Single.fromCallable(() -> {
+                    final String scopes = "oauth2:" + DriveScopes.DRIVE_APPDATA + " " + DriveScopes.DRIVE_FILE;
+                    return GoogleAuthUtil.getToken(context, signInAccount.getAccount(), scopes);
+                }).onErrorReturn(throwable -> {
```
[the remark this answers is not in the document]
HeinousGames: In your original example on updating this from the AsyncTask (https://github.com/wbaumann/SmartReceiptsLibrary/pull/155#discussion_r313194924), you use the onErrorReturn block, but I wasn't sure if I was implementing it correctly. I changed it to doOnError in the latest commit. Please let me know if it needs to be updated again.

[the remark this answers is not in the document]
wbaumann: Checking now! Thanks :)

app/src/main/java/co/smartreceipts/android/sync/drive/DriveServiceHelper.java
```diff
@@ -0,0 +1,197 @@
...
+   * Creates a file in the user's application folder and returns it.
+   */
+  //TODO: Determine if we need the entire file returned, if not, we can call setFields()
+  //      on the execution to tell the api what data of the created file to return
+  public Single<File> createFile(String name, String mimeType, String description,
+                                 Map<String, String> properties, String folderDestination,
+                                 java.io.File javaFile) {
+    return Single.fromCallable(() -> {
+      File metadata = new File();
+      if (!TextUtils.isEmpty(folderDestination))
```
wbaumann: Let's wrap all of these with braces like:

```
if (!TextUtils.isEmpty(folderDestination)) {
    metadata.setParents(Collections.singletonList(folderDestination));
}
```

I know you mostly copy-pasted things, but I think this is a good practice here

app/src/main/java/co/smartreceipts/android/sync/drive/DriveServiceHelper.java
```diff
@@ -0,0 +1,197 @@
...
+    this.context = Preconditions.checkNotNull(context);
+    mDriveService = Preconditions.checkNotNull(driveService);
+  }
+
+  /**
+   * Creates a file in the user's application folder and returns it.
+   */
+  //TODO: Determine if we need the entire file returned, if not, we can call setFields()
+  //      on the execution to tell the api what data of the created file to return
+  public Single<File> createFile(String name, String mimeType, String description,
```
wbaumann: Based on the `isEmpty` checks below, can you make these all (or mostly) `@Nullable`

app/src/main/java/co/smartreceipts/android/sync/drive/GoogleDriveBackupManager.java
```diff
@@ -230,68 +322,73 @@ public Date getLastDatabaseSyncTime() {
...
+    @SuppressLint("CheckResult")
+    private void onGoogleSignInAccountReady(@NonNull GoogleSignInAccount signInAccount) {
+        // First, confirm that we're still initializing
+        synchronized (initializationLock) {
+            if (isInitializing.get()) {
+                Single.fromCallable(() -> {
+                    final String scopes = "oauth2:" + DriveScopes.DRIVE_APPDATA + " " + DriveScopes.DRIVE_FILE;
+                    return GoogleAuthUtil.getToken(context, signInAccount.getAccount(), scopes);
+                }).doOnError(throwable ->  {
+                    Logger.error(GoogleDriveBackupManager.this, "Failed to authenticate user with status: {}", throwable.getMessage());
```
wbaumann: @HeinousGames Looks good now! Before it would return the error string, so it'd use the error string before `GoogleSignInAccountFinalization(signInAccount)`. Now we just log the error 👍

app/src/main/java/co/smartreceipts/android/sync/drive/rx/DriveClientInitializer.kt
```diff
@@ -0,0 +1,78 @@
...
+                             private val googleDriveTableManager: GoogleDriveTableManager,
+                             private val networkManager: NetworkManager) : NetworkStateChangeListener {
+
+    private val compositeDisposable: CompositeDisposable = CompositeDisposable()
+
+    fun initialize() {
+        //TODO: Determine if there is an API update for upload preferences
+        // https://developers.google.com/drive/api/v3/manage-uploads#resumable
+
+        // Monitor for future changes to our wifi settings preference
```
wbaumann: Can we delete all of this? As a general practice, I don't think there's much value to commenting out old code, since it's in the git history if you need to get it

HeinousGames:
Testing took a very long time because of the file's properties. Apparently there are two kinds, properties (visible to any app) and appProperties (only visible to our app). Because all of the files we are saving are in the app folder, my assumption was that we save the device key to the appProperties. Alas, the old implementation, intentional or not, was saving it to the regular properties (begging the question, if the file is in the hidden folder, can another app still see its public properties?). I was checking appProperties, resulting in no device ids being found, so it would create a new backup folder, without a db file (the old one would still be found, and remained in the old folder), breaking most of the functionality in downloading zips of the restore points.

wbaumann:
> Testing took a very long time because of the file's properties. Apparently there are two kinds, properties (visible to any app) and appProperties (only visible to our app). Because all of the files we are saving are in the app folder, my assumption was that we save the device key to the appProperties. Alas, the old implementation, intentional or not, was saving it to the regular properties (begging the question, if the file is in the hidden folder, can another app still see its public properties?). I was checking appProperties, resulting in no device ids being found, so it would create a new backup folder, without a db file (the old one would still be found, and remained in the old folder), breaking most of the functionality in downloading zips of the restore points.

Very interesting. I had no idea :). Did you stick with the old approach of properties?

HeinousGames:
The only way for the new implementation to work correctly with the old one is to continue using the regular properties instead of the app properties. I just didn't realize which one we were using, so any testing that worked prior only worked on fresh installs, and not on devices who upgraded from old to new. Now the issues are resolved so it should be okay to merge.

PR #158, opened 2019-08-21
Dependency updates

HeinousGames:
Things to consider:
- Updating AWS resulted in updating the AmazonS3Client constructor, do we want to be dynamic on the region we pass in?
- Updating Firebase changes the way they handle token generation and messaging. I merged our two services into one, and wasn't sure about how to test this as I've never personally seen notifications from the Smart Receipts app.
- There is a much higher version of Android Chart available, but it still doesn't include your fix for the NPE. Should we open a pull request to have them add it to their code base?
- All tests passed successfully, but many had to be updated in accordance to okhttp, robolectric, mockito kotlin, and aws being updated


gradle.properties
```diff
@@ -1,57 +1,60 @@
...
-PDFBOX_ANDROID_VERSION=1.8.9.1
-PDFIUM_VERSION=1.8.2
+PDFBOX_ANDROID_VERSION=1.8.10.1
+PDFIUM_VERSION=1.9.0
 
 # Note: Periodically check for updates here so we can revert to the official version
 # We want to eventually consume v3.0.4 or higher
-MP_ANDROID_CHART_VERSION=v3.0.2
+# Version 3.1.0 is higher than the waiting to consume version, but still doesn't include Will's fix
+MP_ANDROID_CHART_VERSION=v3.1.0
```
wbaumann: @HeinousGames - Let's revert this to use my version for now. I forked theirs originally as they never bothered to fix that bug. I'm assuming the issue is still open?

wbaumann:
Nice work, @HeinousGames . For your questions:

> Updating AWS resulted in updating the AmazonS3Client constructor, do we want to be dynamic on the region we pass in?

I think we can stick to `US_EAST_1`

> Updating Firebase changes the way they handle token generation and messaging. I merged our two services into one, and wasn't sure about how to test this as I've never personally seen notifications from the Smart Receipts app.

This is for the OCR scans. Ping me, and I'll get a bunch added to your account for testing

> There is a much higher version of Android Chart available, but it still doesn't include your fix for the NPE. Should we open a pull request to have them add it to their code base?

Let's stick with mine. Theirs caused our top crash by a large margin for a while and it was clear they weren't active about fixing it (open for 6+ months with quite a few likes).

> All tests passed successfully, but many had to be updated in accordance to okhttp, robolectric, mockito kotlin, and aws being updated

Sounds good. Thanks for taking this all together :)

gradle.properties
HeinousGames: I couldn't find the issue in their issues section, I wonder if they closed it? I updated the number in our gradle properties, but the library we use is still the same one you made. At the very least, we should pull in their latest changes to our library so we can still have the NPE bug fix with the latest features.

wbaumann: Here was the original issue:

https://github.com/PhilJay/MPAndroidChart/pull/4019

It sounds like it was fixed in a more generic "fix all" commit here:

https://github.com/PhilJay/MPAndroidChart/pull/4019#issuecomment-433469474

So to be honest, I have no idea what their current state is vs mine haha. Let's do this:

1. Keep this as is with their `3.1.0` version, since it's more maintained than mine
2. If any crashes arise in Firebase, we'll revert to mine. 

Sound good?

wbaumann:
@HeinousGames - there's one merge conflict that's preventing the merge. Could you resolve locally? I want to make sure that I don't undo one of your changes. Thanks!

PR #160, opened 2019-08-28
Report total column

HeinousGames:
Decided to leave the row item blank and just display total with tax once in the footer row. Let me know if you think it should be displayed differently.


app/src/main/java/co/smartreceipts/android/model/impl/columns/receipts/ReportTotalColumn.kt
```diff
@@ -0,0 +1,40 @@
...
+        ) {
+
+    override fun getValue(rowItem: Receipt): String = ""
+
+    override fun getFooter(rows: List<Receipt>): String {
+        return if (rows.isNotEmpty()) {
+            val tripCurrency = rows[0].trip.tripCurrency
+            val prices = ArrayList<Price>()
+            prices.add(rows[0].trip.price)
+            for (receipt in rows) {
```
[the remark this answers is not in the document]
HeinousGames: I had previously updated the text from "Report Total" to "Report Total Plus Tax." Now with logic that checks the pre-tax preference, should we have the same logic that displays one string versus the other, or just default to one?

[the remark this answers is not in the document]
wbaumann: Let's keep the title always as "Report Total" but include the correct result based on the user's preference

app/src/main/java/co/smartreceipts/android/model/impl/columns/receipts/ReportTotalColumn.kt
```diff
@@ -0,0 +1,45 @@
...
+                id,
+                ReceiptColumnDefinitions.ActualDefinition.REPORT_TOTAL,
+                syncState,
+                customOrderId,
+                uuid
+        ) {
+
+    override fun getValue(rowItem: Receipt): String = ""
+
+    override fun getFooter(rows: List<Receipt>): String {
```
wbaumann: Actually - I'm realizing that this also won't fully work, since it omits the distance total...

The more I look at this, the more that I realize that it's going to be confusing for our customers to have this addition (vs the top of the report). I think it may be better to discard this altogether and stick with the version at the top of the report.

Sorry for making you go through the work with this but does that work?

HeinousGames: That's okay Will. It's all in a separate branch anyways so we can just delete the branch and forget about it. The report total is already displayed at the top, was there anything we needed to update regarding that?

wbaumann: Na - I think that should be good. Sorry for the false alarm but thanks for working with me on this!

PR #162, opened 2019-09-06
Close quietly deprecation

HeinousGames:
While researching the pdf import bug, I noticed that IOUtils deprecated their closeQuietly method. Recommendations were to replace with try-with-resources, so I updated them on this branch. Should be noted that there are three closeQuietly's still existing as I wasn't really sure on how to refactor those remainders. They can be found in PDFGridRenderer, LollipopPdfPDImageXFactory, and CompatPdfPDImageXFactory class files.


wbaumann:
@HeinousGames I merged in the other two changes, but this led to a conflict on this one. I think you'll have to merge in `master` to this branch first :)

PR #165, opened 2019-09-17
Admob id update

HeinousGames:


wbaumann:
@HeinousGames - No conflicts but looks like there's a build failure for this one

app/src/main/AndroidManifest.xml
```diff
@@ -179,6 +168,10 @@
             android:name="google_analytics_adid_collection_enabled"
             android:value="false" />
 
+        <meta-data
+            android:name="com.google.android.gms.ads.APPLICATION_ID"
+            android:value="@string/admob_ad_id"/>
```
wbaumann: @HeinousGames Here's the issue. This is in our main manifest, which means all variants will attempt to use it.

Can you add this instead to the `app/src/free/AndroidManifest.xml`. Then it should work

PR #171, opened 2019-09-28
Bugfixes

JuliaSoboleva:
* The app crashed while it was creating PDF report if there were PNG files (`ImageImportProcessor` saved all images as JPEG files)

* The app couldn't create PDF report correctly if there were cropped images (it was my error, SmartCropper is good)

All fixed :)


JuliaSoboleva:
@wbaumann No, it won't fix this issue for cropped images. I'll write some code to check and update such images during the pdf construction process. I think tomorrow it will be done.

PR #172, opened 2019-09-28
Google drive api limit reached

HeinousGames:
Not sure if this fix is adequate enough to get rid of the api limit exceeded bugs, we may have to put it in beta and see how it handles. We can also add the oauth token to our api calls, but I didn't see any clear indication that both the key and user token were required.


app/src/main/java/co/smartreceipts/android/sync/drive/DriveServiceHelper.java
```diff
@@ -127,25 +131,30 @@
...
     return Single.fromCallable(() ->
-            mDriveService.files().list().setSpaces(DRIVE_SEARCH_AREA).setOrderBy("modifiedTime").setFields("*").execute());
+            mDriveService.files().list().setSpaces(DRIVE_SEARCH_AREA).setOrderBy("modifiedTime").setFields("*")
+                    .setKey(context.getString(R.string.android_api_key)).execute());
   }
 
   public Single<java.io.File> getDriveFileAsJavaFile(String fileId, java.io.File downloadLocationFile) {
-    return Single.fromCallable(() -> mDriveService.files().get(fileId).setFields("*").executeMediaAsInputStream())
+    return Single.fromCallable(() -> mDriveService.files().get(fileId).setFields("*")
+            .setKey(context.getString(R.string.android_api_key)).executeMediaAsInputStream())
```
wbaumann: Where is this key from? And is it their public one or private one? Even though we've encrypted it, you can determine the value from the APK, so I just want to make sure it's a lower risk item

HeinousGames: I located the key from the google apis console (https://console.developers.google.com/apis/credentials?project=smart-receipts&supportedpurview=project). It is the Android key specifically. I believe it is also located in the google-services.json file we use for firebase, not sure if that file is meant to be encrypted or not.

wbaumann:
@HeinousGames - It looks like the build is failing here too. I'd like to fix this before we merge in the changes

PR #173, opened 2019-09-29
Added code that fixes consequences of a bug with cropped images

JuliaSoboleva:
I've added code that will fix all "broken" cropped image files.

We will need to remove this code after some time. 

I thought it would be good to add new analytics event like "broken image was fixed" for you to check when it won't be used anymore. I decided not to add a new dependency on analytics in this large `EmailAssistant` class yet. But if you want, we can quickly add it, just let me know.


wbaumann:
This works great. Thanks @JuliaSoboleva

wbaumann:
I think we'll need to refactor the `EmailAssistant` at a later date, so no worries on that now

PR #175, opened 2019-10-01
Read permission fix

HeinousGames:
Asks user for storage permission. If they accept it will work as intended and not ask again (unless the app forgets the preference or if the user changes it). If the user denies the request, we show an alert letting them know they must accept to import the file, at which point they can click import again and then accept. If they deny and mark never ask again, we alert them that they must manually change the permission in the setting and get them to the app settings page where they can change it.


app/src/main/java/co/smartreceipts/android/sync/widget/backups/BackupsFragment.java
```diff
@@ -109,13 +118,17 @@ public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup c
...
-            final Intent intent = new Intent(Intent.ACTION_GET_CONTENT);
-            intent.addCategory(Intent.CATEGORY_OPENABLE);
-            intent.setType("*/*");
-            try {
-                startActivityForResult(Intent.createChooser(intent, getString(R.string.import_string)), IMPORT_SMR_REQUEST_CODE);
-            } catch (android.content.ActivityNotFoundException ex) {
-                Toast.makeText(getContext(), getString(R.string.error_no_file_intent_dialog_title), Toast.LENGTH_SHORT).show();
+
+            // ask permission first
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
```
wbaumann: Check out the flow for how this works for the PDF imports. It only asks for permissions if it's a `file://` type item. If it's a `context://` one, you don't need to ask for permission (all newer requests should use the context type, but it's not 100% true)

app/src/main/java/co/smartreceipts/android/sync/widget/backups/BackupsFragment.java
```diff
@@ -208,6 +221,66 @@ public void onProviderChanged(@NonNull SyncProvider newProvider) {
...
+                }
+            }
+        }
+    }
+
+    private void alertAboutPermissions() {
+        final AlertDialog.Builder dlgAlert = new AlertDialog.Builder(getActivity());
+        dlgAlert.setCancelable(false)
+                .setIcon(R.mipmap.ic_launcher)
+                .setTitle("Storage Permission Required")
```
wbaumann: All strings should live in the `R.string` (`strings.xml`) file, so we can translate them... But you should be able to find already translated ones there

app/src/main/java/co/smartreceipts/android/sync/widget/backups/BackupsFragment.java
```diff
@@ -175,7 +185,20 @@ public void onActivityResult(int requestCode, int resultCode, @Nullable Intent d
         if (resultCode == Activity.RESULT_OK) {
             if (requestCode == IMPORT_SMR_REQUEST_CODE) {
                 if (data != null) {
-                    navigationHandler.showDialog(ImportLocalBackupDialogFragment.newInstance(data.getData()));
+                    uri = data.getData();
+                    if (uri.getScheme().equals("file")) {
```
wbaumann: Is it possible to share these import flows with the other parts that need them? So we don't duplicate the code here? 

Cool if not but wanted to check

HeinousGames: Are you saying to extract the request permission methods into a singleton/constants type class where we can access them for any permission needed by the user? I've done this for other apps, so I can do it here too, I just need to verify that is what you are asking for. Is there already such a class with public static methods that any area of the app can access?

wbaumann: Yea sorry for the lack of clarity. The longer version is that we have this class already:

https://github.com/wbaumann/SmartReceiptsLibrary/blob/8682d812be4a439ff44a8e04becdbcaedd1b1e91/app/src/main/java/co/smartreceipts/android/imports/intents/widget/info/IntentImportInformationInteractor.java#L26

Which we use for the PDF import flow. I'd like to use that same of chunk for this code instead, so we don't duplicate the logic between these two places

HeinousGames: I'm sorry, but I'm wildly confused as to how this code works. I'm not sure whether I'm supposed to use the interactor or the presenter, and I can't seem to find a clear example of them being used elsewhere in the code base to refer to.

I noticed that the Interactor has a process method that should take an intent and return whether or not I should be allowed to interact with that data based on the permission being granted or not. When I try passing the intent data in, I'm not sure what's supposed to come out. I can flatMap to get a UIIndicator of the intent, but I just run through the same logic with it that the process method does, so it seems redundant, and expects me to return something further after the fact.

The presenter class has a subscribe method that seems to call the interactor's process method, but I have no idea how the subscribe method gets the returned intent from the fragment. I never pass it into the subscribe method, so how do I know it's looking at the intent at the right time?

HeinousGames:
Instead of making a new presenter that works with Instrumentation.ActivityResult class, I was able to get the current presenter to work with the fragment by setting the SmartReceiptsActivity's intent as the one coming back from the result. Passing in a normal file that doesn't require permissions failed though, so I had to add an application/x-zip mime type to our supported mime type list.

app/src/main/java/co/smartreceipts/android/activities/SmartReceiptsActivity.java
```diff
@@ -283,7 +288,43 @@ public void presentIntentImportInformation(@NonNull FileType fileType) {
...
+        final AlertDialog.Builder dlgAlert = new AlertDialog.Builder(this);
+        dlgAlert.setCancelable(false)
+                .setIcon(R.mipmap.ic_launcher)
+                .setTitle(getString(R.string.storage_permission_required))
+                .setMessage(getString(R.string.permission_must_be_granted))
+                .setNeutralButton(getString(R.string.ok), null).show();
+    }
+
+    private void alertChangePermissions(final Context context) {
+        final AlertDialog.Builder dlgAlert = new AlertDialog.Builder(context);
```
wbaumann: Let's change this to instead use a `DialogFragment`. It's mostly the same thing but in fragment form.

The main benefit is that it handles rotation (if you get an AlertDialog and rotate, it'll automatically dismiss itself)

app/src/main/java/co/smartreceipts/android/activities/SmartReceiptsActivity.java
```diff
@@ -283,7 +289,16 @@ public void presentIntentImportInformation(@NonNull FileType fileType) {
 
     @Override
     public void presentIntentImportFatalError() {
-        Toast.makeText(this, R.string.toast_attachment_error, Toast.LENGTH_SHORT).show();
-        finish();
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+            Bundle args = new Bundle();
+            args.putBoolean("shouldShowRationale", shouldShowRequestPermissionRationale(Manifest.permission.READ_EXTERNAL_STORAGE));
+            PermissionAlertDialogFragment permissionDialogFragment = PermissionAlertDialogFragment.newInstance();
```
wbaumann: Let's move the `bundle` stuff into the `newInstance()` method like:

```
... newInstance(String permission) {
Bundle args = new Bundle();
args.putBoolean(...)
}
```

So the logic is contained entirely in that fragment

app/src/main/java/co/smartreceipts/android/fragments/PermissionAlertDialogFragment.java
```diff
@@ -0,0 +1,53 @@
...
+    @Override
+    public Dialog onCreateDialog(Bundle savedInstanceState) {
+        AlertDialog.Builder builder = new AlertDialog.Builder(getContext())
+                .setCancelable(false)
+                .setTitle(getString(R.string.storage_permission_required));
+
+        if (getArguments().getBoolean("shouldShowRationale")) {
+            builder.setIcon(R.mipmap.ic_launcher)
+                    .setMessage(getString(R.string.permission_must_be_granted))
+                    .setNeutralButton(getString(R.string.ok), null);
```
wbaumann: Let's use `setPositiveButton` here instead. There's no special motivation for it, but I think Google styles the positive one with the primary colors

app/src/main/java/co/smartreceipts/android/fragments/PermissionAlertDialogFragment.java
```diff
@@ -0,0 +1,61 @@
...
+import androidx.fragment.app.DialogFragment;
+
+import co.smartreceipts.android.R;
+
+/**
+ * Dialog Fragment which asks if user wants to leave feedback
+ */
+public class PermissionAlertDialogFragment extends DialogFragment {
+
+    @SuppressLint("NewApi")
```
wbaumann: One final question - Why do we need the `@SuppressLint("NewApi")` method?

HeinousGames: I guess that wasn't the best tag to choose. We already do a build version check before calling the fragment, so I changed the tag to requires Android M.

PR #176, opened 2019-10-03
Kt lint

HeinousGames:
- KTLint may have to be installed on your machine to have the new gradle tasks run.
- You may also have to [apply ktlint](https://github.com/pinterest/ktlint#option-2) to your project locally. At which point you can choose whether you want to follow Android's kotlin code style or not.
- Running ./gradlew ktlint will show which files are not following which rules.
- Running ./gradlew ktlintFormat will correct as many of the errors as possible.
- We can also add ktlint to [CircleCI](https://github.com/sanogueralorenzo/Android-Kotlin-Clean-Architecture/pull/33/files) to correct code before it compiles and builds.


wbaumann:
@HeinousGames - You mention:

> KTLint may have to be installed on your machine to have the new gradle tasks run.

Is there a way instead to include this in the `./gradlew test` script (ideally as an extension of the `./gradlew clean check`), so it happens automatically?

HeinousGames:
using the new gradle plugin, the ktlintCheck tasks are added to the gradle check task automatically. Each one exits with a failure of non-zero, which is normal, so when running the normal check task, you must add the --continue tag to it so it runs the entire way through.

wbaumann:
So to be clear for this:

> using the new gradle plugin, the ktlintCheck tasks are added to the gradle check task automatically. Each one exits with a failure of non-zero, which is normal, so when running the normal check task, you must add the --continue tag to it so it runs the entire way through.

The `check` task will abort if KLint fails, correct? And you can override it with the `--continue` flag?

wbaumann:
Also - does it currently pass?

HeinousGames:
Yes, running `./gradlew check` will fail as soon as the first ktlint check task exits with its non-zero failure. Running `./gradlew check --continue` will allow the check command to continue running, and each ktlint check task will say it "failed," despite that being how it works. The check task takes a long time to run, and after I ran it, it only showed failures for the ktlint check tasks that ran as apart of it. I didn't see any errors from unit tests, and adding these ktlint check tasks might make travis or circleci builds take longer if they run said check task.

KTLint's check tasks only check the code, and show you what changes you can make. The ktlint format tasks will automatic refactor the code so they are resolved of the warnings that can pop up in the check tasks. I am not running the format tasks as it will change almost all of our kotlin files, and I think you'd prefer to choose your ruleset before we utilize the format task.

wbaumann:
Sounds good! I think we should try to lock in some standard formatting behaviors in a future change, but this one is good to start :)

PR #177, opened 2019-10-05
Google drive api token adjustment

HeinousGames:
I think this will work. I still didn't see any errors, and I'm guessing we still won't see it unless running a production version, just like with the last attempted change. Although I saw one site saying to use keys or tokens, I saw a separate saying only use tokens. I guess I shouldn't have added that last change, my mistake. Let's hope this resolves the issue!


app/src/main/java/co/smartreceipts/android/sync/drive/GoogleDriveBackupManager.java
```diff
@@ -332,7 +332,7 @@ private void onGoogleSignInAccountReady(@NonNull GoogleSignInAccount signInAccou
                 }).doOnError(throwable ->  {
                     Logger.error(GoogleDriveBackupManager.this, "Failed to authenticate user with status: {}", throwable.getMessage());
                 }).subscribeOn(Schedulers.io())
-                        .subscribe(onSuccess -> GoogleSignInAccountFinalization(signInAccount),
+                        .subscribe(onSuccess -> GoogleSignInAccountFinalization(signInAccount, onSuccess),
```
wbaumann: Can we rename `onSuccess` to `token` here?

PR #181, opened 2019-11-01
Pre-filled Name field

mascygp:
Can you add a setting option to have the field "Name" already filled when I add a new recepit.

Settings -> Receipts -> Name = "Minnie"
When I add a new receipt, the fiel Name already contains the string Minnie

pre-filled-Name


wbaumann:
So to clarify, you want all of your receipts to always have the same name, `Minnie`? And you'd configure that in the settings, so it always has the same value?

PR #182, opened 2019-11-04
Hash column

HeinousGames:


app/src/main/java/co/smartreceipts/android/persistence/database/tables/adapters/ReceiptDatabaseAdapter.java
```diff
@@ -105,6 +106,7 @@ public Receipt readForSelection(@NonNull Cursor cursor, @NonNull Trip trip, bool
         final String currency = cursor.getString(currencyIndex);
         final boolean fullPage = !(cursor.getInt(fullPageIndex) > 0);
         final int paymentMethodId = cursor.getInt(paymentMethodIdIndex);
+        final String imgHash = (imgHashIndex > 0) ? cursor.getString(imgHashIndex) : "";
```
wbaumann: I think we can let this value be `null` instead if there's no image

app/src/main/java/co/smartreceipts/android/activities/NavigationHandler.java
```diff
@@ -83,11 +83,11 @@ public void navigateToReportInfoFragment(@NonNull Trip trip) {
         }
     }
 
-    public void navigateToCreateNewReceiptFragment(@NonNull Trip trip, @Nullable File file, @Nullable OcrResponse ocrResponse) {
+    public void navigateToCreateNewReceiptFragment(@NonNull Trip trip, @Nullable File file, @Nullable OcrResponse ocrResponse, Boolean shouldGenerateHash) {
         if (isDualPane) {
-            replaceFragmentWithAnimation(fragmentProvider.newCreateReceiptFragment(trip, file, ocrResponse), R.id.content_details, R.anim.enter_from_bottom, DO_NOT_ANIM);
+            replaceFragmentWithAnimation(fragmentProvider.newCreateReceiptFragment(trip, file, ocrResponse, shouldGenerateHash), R.id.content_details, R.anim.enter_from_bottom, DO_NOT_ANIM);
```
wbaumann: For all of these, why do we use the `shouldGenerateHash`? We can always generate it based on if the file != null

app/src/main/java/co/smartreceipts/android/model/converters/DistanceToReceiptsConverter.java
```diff
@@ -99,6 +99,7 @@ private Receipt generateReceipt(@NonNull List<Distance> distancesThisDay) {
         factory.setTrip(distance0.getTrip());
         factory.setDate(distance0.getDate());
         factory.setFile(null);
+        factory.setImgHash("");
```
wbaumann: Let's set to `null` instead (to better align with the optional

app/src/main/java/co/smartreceipts/android/receipts/editor/ReceiptCreateEditFragmentPresenter.java
```diff
@@ -149,6 +153,10 @@ void saveReceipt(Date date, TimeZone timeZone, String price, String tax,
                 .setExtraEditText2(extraText2)
                 .setExtraEditText3(extraText3);
                 // Note: We don't set the custom_order_id. This happens in the ReceiptTableActionAlterations
+        
+        if (fragment.shouldGenerateHash()) {
```
wbaumann: We should do this in a background process to avoid blocking the UI thread.

app/src/main/java/co/smartreceipts/android/activities/NavigationHandler.java
wbaumann: Or is this specific to checking if the image was taken by photo vs imported

HeinousGames: Yes, I'm trying to make sure that we only generate a hash on an image file if the photo came from the camera, or is cropped from the camera import. Generating for any file could result in  generating a modified picture that was added to a device and imported from the gallery.

HeinousGames:
> @HeinousGames This is unrelated to the current approach, but I'm thinking that using this "sometimes we have the MD5, sometimes we don't approach" will end up being more for users. What do you think if we do the following instead:
> 
> * Add a new column (entitled Image Hash - SHA256)
> * Change the hashing method to SHA256: https://www.baeldung.com/sha-256-hashing-java
> * Rather than saving this in the database, let's just generate it at runtime (then we won't need a new database version)
> 
> Does that work for you?

Confused about how to generate the hash at runtime. Are you saying that we generate a hash for the receipt when we add it to the app, or generate hashes for the images when we generate the report? If it is during report generation, don't we still need a bool to refer to whether the picture file was added via the camera option or importing. Can we persist this boolean without using the database?

app/src/main/java/co/smartreceipts/android/receipts/editor/ReceiptCreateEditFragmentPresenter.java
HeinousGames: How is this blocking the UI thread? Right below it we do something similar in the insert method, calling fragment.getFile(). `builderFactory.setFile(fragment.getFile())`
Is it because that fragment call is within the insert method?

HeinousGames:
I made some changes, but will remove the database changes after you answer my questions above regarding persistence. Also, after updating to SHA256 the hash column is so wide that I can only have 3 columns in portrait mode, is there a better way to display the value in the table?

wbaumann:
> > @HeinousGames This is unrelated to the current approach, but I'm thinking that using this "sometimes we have the MD5, sometimes we don't approach" will end up being more for users. What do you think if we do the following instead:
> > 
> > * Add a new column (entitled Image Hash - SHA256)
> > * Change the hashing method to SHA256: https://www.baeldung.com/sha-256-hashing-java
> > * Rather than saving this in the database, let's just generate it at runtime (then we won't need a new database version)
> > 
> > Does that work for you?
> 
> @wbaumann Confused about how to generate the hash at runtime. Are you saying that we generate a hash for the receipt when we add it to the app, or generate hashes for the images when we generate the report? If it is during report generation, don't we still need a bool to refer to whether the picture file was added via the camera option or importing. Can we persist this boolean without using the database?

I'm thinking that we simplify things and ONLY generate it on runtime (i.e. for the column). I know it doesn't strictly deal with the original requirements, but I'd rather start with this and see if it's actually sufficient (I'm not familiar enough with the nuances, but it seems like it should cover the case)

app/src/main/java/co/smartreceipts/android/receipts/editor/ReceiptCreateEditFragmentPresenter.java
wbaumann: Ah missed that. My bad.

Both of those `getFile()` calls are blocking, but it's fine if they're already of the UI thread

HeinousGames:
@wbaumann I have removed the db logic and now image hashes are generated at runtime for all images. They always generate the same in pdf and csv. Is there something we can do about changing how it takes up space on the generated reports? I can only fit max 3 columns when I include the hash... or should it only be used in a separate report filing?

wbaumann:
@HeinousGames Didn't even think of that. Maybe we can improve it with a different algorithm like in this discussion:

https://stackoverflow.com/questions/4567089/hash-function-that-produces-short-hashes

Adler-32 doesn't seem especially tamper proof, but it looks to produce short results

app/src/main/java/co/smartreceipts/android/model/impl/columns/receipts/ReceiptFileHashColumn.kt
```diff
@@ -0,0 +1,59 @@
...
+                var numRead = 0
+                while (numRead != -1) {
+                    numRead = inputStream.read(buffer)
+                    if (numRead > 0)
+                        digest.update(buffer, 0, numRead)
+                }
+                val hashedBytes = digest.digest()
+                return convertHashToString(hashedBytes)
+            }
+        } catch (e: Exception) {
```
wbaumann: Let's only catch the explicit exceptions like IOException or whatever instead of all. It's usually a bad practice, since it can mask NullPointerExceptions, etc

HeinousGames: @wbaumann Looks like Adler himself says that the Adler32 method shouldn't be used to get a hash: https://stackoverflow.com/questions/11597762/adler32-repeating-very-quickly

Not sure if this applies to our scenario or not. You link also mentioned truncating the original hash value that's returned, but is that safe?

wbaumann: Good find. I think that's more of a concern in "high security" environments.

Let's stick with your current (SHA) approach then and only take the last `X` digits (8? 16? Whatever looks best on the result)

app/src/main/java/co/smartreceipts/android/model/impl/columns/receipts/ReceiptFileHashColumn.kt
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+    }
+
+    private fun convertHashToString(hashedBytes: ByteArray): String {
+        val returnVal = StringBuilder()
+        for (hashedByte in hashedBytes) {
+            returnVal.append(((hashedByte and 0xff) + 0x100).toString(16).substring(1))
+        }
+        return returnVal.toString().substring(returnVal.lastIndex-16, returnVal.lastIndex).toUpperCase(Locale.getDefault())
```
wbaumann: Let's change this to take advantage of some Kotlin extensions that aren't present in Java: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/take-last.html

```
returnVal.takeLast(16).toUpperCase(Locale.EN)
```

Also, let's use `Locale.EN` to ensure that we're not doing any weird logic (normally you don't want to hardcode the Locale, but a Hash Digest will use ASCII chars, so we should keep it to English rules)

app/src/main/java/co/smartreceipts/android/model/impl/columns/receipts/ReceiptFileHashColumn.kt
```diff
@@ -0,0 +1,60 @@
...
+        } catch (e: NoSuchAlgorithmException) {
+            return ""
+        }
+
+    }
+
+    private fun convertHashToString(hashedBytes: ByteArray): String {
+        val returnVal = StringBuilder()
+        for (hashedByte in hashedBytes) {
+            returnVal.append(((hashedByte and 0xff) + 0x100).toString(16).substring(1))
```
wbaumann: What's the `.substring(1)` do here?

HeinousGames: I'm actually not sure why it's there, but I do know that a lot of hashing examples online include it: https://stackoverflow.com/questions/9661008/compute-sha256-hash-in-android-java-and-c-sharp

After some light reading, my understanding is that when the value is returned it will include a hash symbol (#) as the first character: https://forum.bugcrowd.com/t/what-does-this-js-code-do-and-how-to-exploit/1998/2

PR #184, opened 2019-11-14
Search feature

JuliaSoboleva:
Implemented search feature. Now users can perform a search
- for trips (by name, comment)
- for receipts (by name, comment, price, category, payment method)

I added [AdapterDelegates](https://github.com/sockeqwe/AdapterDelegates) library in a new `SearchActivity` for better and clearer code. I hope to reuse adapter delegates for older screens a bit later.

Unfortunately, I didn't implement searching by date yet. I need to investigate it a bit more. 

And I didn't implement highlights in the search results. Since we're performing searching by comments and payment methods (that are not visible on the standard item card layouts), I think that would be good to show extended versions of receipt's and trip's layouts which will include all information. This way we can avoid the case when the item is present in the search results but there are no visible search matching and no highlights. What do you think about this? :)


app/src/main/java/co/smartreceipts/android/search/SearchInteractor.kt
```diff
@@ -0,0 +1,119 @@
...
+        fun isEmpty() = trips.isEmpty() && receipts.isEmpty()
+    }
+
+
+    @Inject
+    constructor(databaseHelper: DatabaseHelper) : this(databaseHelper, Schedulers.io(), AndroidSchedulers.mainThread())
+
+    fun getSearchResults(input: String): Single<SearchResults> {
+
+        // TODO: 10.11.2019 investigate how we can perform search by date
```
wbaumann: Honestly - we can probably just skip this for now. I doubt it'll be a common use case

app/src/main/java/co/smartreceipts/android/persistence/DatabaseHelper.java
```diff
@@ -303,6 +306,50 @@ private void queryTripDailyPrice(final Trip trip) {
         return mFullCurrencyList;
     }
 
+    public List<String> search(@NonNull String input, @Nonnull String tableName, @Nonnull String resultColumn,
```
wbaumann: Can we wrap this in a `Single` to help avoid the risk of calling this on the UI thread

app/src/main/java/co/smartreceipts/android/search/SearchInteractor.kt
```diff
@@ -0,0 +1,119 @@
...
+        }
+
+        // search by category name and code
+        val categoriesList = databaseHelper.search(
+            input, CategoriesTable.TABLE_NAME, CategoriesTable.COLUMN_ID, null,
+            CategoriesTable.COLUMN_NAME, CategoriesTable.COLUMN_CODE
+        )
+
+        // search by payment method name
+        val paymentMethodsList = databaseHelper.search(
```
wbaumann: Let's replace each of these with a `Single` and then combine/zip them all up. I'm not sure that these calls are off the UI thread, since we don't use the scheduler until a later block of code

wbaumann:
Awesome work with this change. Really excited to have it live :)

For your notes:

> I added AdapterDelegates library in a new SearchActivity for better and clearer code. I hope to reuse adapter delegates for older screens a bit later.

Perfect. Sounds like a good change in an upcoming PR. Whenever you want is cool

> Unfortunately, I didn't implement searching by date yet. I need to investigate it a bit more.

No worries. We can skip this!

> And I didn't implement highlights in the search results. Since we're performing searching by comments and payment methods (that are not visible on the standard item card layouts), I think that would be good to show extended versions of receipt's and trip's layouts which will include all information. This way we can avoid the case when the item is present in the search results but there are no visible search matching and no highlights. What do you think about this? :)

I think that's a reasonable approach. I'm working with a UX designer right now to help improve some of the layouts. Let me include that in this discussion and get some thoughts on that. I think it'll help. Overall - your current solution seems great, but I agree there's probably a way to further improve the UX here

JuliaSoboleva:
@wbaumann Fixed :)

wbaumann:
Thanks @JuliaSoboleva !

PR #186, opened 2019-12-01
Distance payment method

HeinousGames:


app/src/main/java/co/smartreceipts/android/distance/editor/DistanceCreateEditFragment.kt
```diff
@@ -156,15 +193,47 @@ class DistanceCreateEditFragment : WBFragment(), DistanceCreateEditView, View.On
         }
     }
 
+    override fun onActivityCreated(savedInstanceState: Bundle?) {
+        super.onActivityCreated(savedInstanceState)
+
+        paymentMethodTableEventsListener = object : StubTableEventsListener<PaymentMethod>() {
+            override fun onGetSuccess(list: List<PaymentMethod>) {
+                if (isAdded) {
+                    // TODO: Move to payment methods presenter
```
wbaumann: Can we move this all to the presenter? That way we can add test that this logic actually works

HeinousGames: @wbaumann I'm failing to understand how to make this change. Originally I copied this implementation from the receipt create edit fragment (that version also has the todo for moving it to the payment methods presenter). Which one am I trying to move to the presenter? Is it the isAdded logic, the onGetSuccess logic, or creating the entire listener and subscribing it? The table events listener requires the payment methods table controller to subscribe it, do I want to move that object to the presenter as well?

wbaumann: So the idea here is that every time we view this screen, we do the following:

1. When `onResume` is called, we attempt to fetch all payment methods from that DB
2. When we get a result, we add `PaymentMethod.NONE`
3. Finally, we attempt to determine the "selected" item

At the moment, all this logic exists in the Fragment, which means there's no easy way to test it independently of "Android". My proposal is creating presenter, which does the following in it's `subscribe()` method:

1. When `onResume` is called, we attempt to fetch all payment methods from that DB
2. When we get a result, we add `PaymentMethod.NONE`
3. Calls `view.displayPaymentMethods(...)`

And the fragment can then consume that view.

Does that make sense?

HeinousGames: @wbaumann I think I figured it out, let me know if it can be better.

app/src/main/java/co/smartreceipts/android/distance/editor/DistanceCreateEditFragment.kt
```diff
@@ -264,6 +305,35 @@ class DistanceCreateEditFragment : WBFragment(), DistanceCreateEditView, View.On
...
+            } else {
+                ViewCollections.run(paymentMethodsViewsList, ButterKnifeActions.setVisibility(View.GONE))
+            }
+        }
+    }
+
+    override fun displayPaymentMethods(list: List<PaymentMethod>) {
+        if (isAdded) {
+            val paymentMethods = ArrayList(list)
+            paymentMethods.add(PaymentMethod.NONE)
```
wbaumann: Can we also add this to the `List<>` that gets built in the presenter? I mention this, since we duplicate it in both places that we use this :)

app/src/test/java/co/smartreceipts/android/receipts/editor/paymentmethods/PaymentMethodsPresenterTest.kt
```diff
@@ -42,23 +46,23 @@ class PaymentMethodsPresenterTest  {
     @Test
     fun subscribeWithPaymentMethodsEnabled() {
         whenever(userPreferenceManager.getSingle(UserPreference.Receipts.UsePaymentMethods)).thenReturn(Single.just(true))
-        val presenter = PaymentMethodsPresenter(view, userPreferenceManager, Schedulers.trampoline(), Schedulers.trampoline())
+        val presenter = PaymentMethodsPresenter(view, userPreferenceManager, Schedulers.trampoline(), Schedulers.trampoline(), paymentMethodsTableController)
```
wbaumann: Let's also add a verification that `view.displayPaymentMethods(...)` gets called when it's enabled (and not when it's disabled)

app/src/main/java/co/smartreceipts/android/distance/editor/DistanceCreateEditFragment.kt
wbaumann: Great work! Exactly what I was thinking. That now helps us to reduce the amount of duplication between the different implementations. Left a few other minor notes below to help improve that further

app/src/test/java/co/smartreceipts/android/receipts/editor/paymentmethods/PaymentMethodsPresenterTest.kt
HeinousGames: Hey @wbaumann I'm not sure how to do this. It seems I can only verify that the togglePaymentMethodFieldVisibility method from the view was called, and not the displayPaymentMethods method. Should I be subscribing the listener to the controller differently (like how the compositeDisposable tracks the preference to call the view's toggle method)? I can only verify that the controller subscribed the listener and the get method was called to make the onGetSuccess launch, but I'm just not sure how to verify the onGetSuccess/view.displayPaymentMethods.

wbaumann: The `displayPaymentMethods` gets called as part of that flow, correct? Here's how I understand that this flow works:

1. You call `subscribe()`
2. Some processing work happens
3. We call `displayPaymentMethods` with data about the current set up payment methods

Is that right? In this case, we should be able to call `verify(view.displayPaymentMethods(...))`. Does that make sense?

HeinousGames: @wbaumann I understood the request, I just couldn't figure out how to actually test it. I had to do some research on mockito and rxjava to better understand, but I think I figured it out. I updated the presenter so that I can verify the method, but I can't verify it's not called when disabled. I believe that no matter what, the controller is called every time the presenter's subscribe method is called, so a verify never test will fail.

wbaumann: For the disabled state, what's the flow? Do we still always call `displayPaymentMethods`?

One way to ignore that (not a big deal here), is to do a check like:

```
if (enabled) {
  view.displayPaymentMethods(...)
}
```

In your presenter. Nothing major as long as it doesn't display thoug

PR #187, opened 2019-12-01
Hide auto complete entry

HeinousGames:


app/src/main/java/co/smartreceipts/android/autocomplete/AutoCompleteArrayAdapter.kt
```diff
@@ -3,14 +3,25 @@ package co.smartreceipts.android.autocomplete
...
  */
 class AutoCompleteArrayAdapter<Type>(context: Context,
-                                     autoCompleteResults: List<AutoCompleteResult<Type>>)
-    : ArrayAdapter<AutoCompleteResult<Type>>(context, android.R.layout.simple_dropdown_item_1line, autoCompleteResults) {
+                                     autoCompleteResults: ArrayList<AutoCompleteResult<Type>>,
+                                     private val listener: ClickListener)
+    : ArrayAdapter<AutoCompleteResult<Type>>(context, R.layout.auto_complete_view, autoCompleteResults) {
+
+    interface ClickListener {
+        fun callback(removeAutoCompleteResult: Boolean, position: Int)
```
wbaumann: Let's name this to `onClick`

app/src/main/java/co/smartreceipts/android/autocomplete/AutoCompleteArrayAdapter.kt
```diff
@@ -27,4 +38,24 @@ class AutoCompleteArrayAdapter<Type>(context: Context,
             super.notifyDataSetChanged()
         }
     }
+
+    override fun getView(position: Int, convertView: View?, parent: ViewGroup): View? {
+        var listItem: View? = convertView
```
wbaumann: Let's make this a `val` and do this in a single line like:

`val listItem = convertView ?: LayoutInflater.from(context).inflate(R.layout.auto_complete_view, parent, false)`

With the "Elvis operator" `?:` (named because the question mark looks like Elvis's hair)

app/src/main/java/co/smartreceipts/android/autocomplete/AutoCompleteInteractor.kt
```diff
@@ -34,43 +37,82 @@ class AutoCompleteInteractor<Type> constructor(private val provider: AutoComplet
...
-                                    if (!resultsSet.contains(displayName)) {
-                                        val result = AutoCompleteResult(displayName, it)
-                                        resultsSet[displayName] = result
-                                        results.add(result)
-                                    } else {
-                                        resultsSet[displayName]!!.additionalItems.add(it)
+                                    // make sure value wasn't removed by user
+                                    var removedByUser = false
+                                    if (it is Receipt) {
+                                        if (field.name() == "Name") {
```
wbaumann: This isn't safe? Does it handle translations into other languages?

Let's try to use an enum and switch it a `when` statement instead to make more readable.

app/src/main/java/co/smartreceipts/android/autocomplete/AutoCompleteView.kt
```diff
@@ -26,6 +26,6 @@ interface AutoCompleteView<Type> {
      * @param results a [List] of [AutoCompleteResult] of [Type]
      */
     @UiThread
-    fun displayAutoCompleteResults(field: AutoCompleteField, results: List<AutoCompleteResult<Type>>)
+    fun displayAutoCompleteResults(field: AutoCompleteField, results: ArrayList<AutoCompleteResult<Type>>)
```
wbaumann: Add `List` here (along with the rest)

app/src/main/java/co/smartreceipts/android/autocomplete/AutoCompleteArrayAdapter.kt
```diff
@@ -27,4 +38,22 @@ class AutoCompleteArrayAdapter<Type>(context: Context,
             super.notifyDataSetChanged()
         }
     }
+
+    override fun getView(position: Int, convertView: View?, parent: ViewGroup): View? {
+        val listItem: View? = convertView ?: LayoutInflater.from(context).inflate(R.layout.auto_complete_view, parent, false)
```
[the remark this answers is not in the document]
HeinousGames: Done!

app/src/main/java/co/smartreceipts/android/autocomplete/AutoCompleteInteractor.kt
```diff
@@ -34,43 +40,65 @@ class AutoCompleteInteractor<Type> constructor(private val provider: AutoComplet
      * @param field the [AutoCompleteField] to use
      * @param input the current user input [CharSequence]
      *
-     * @return a [Maybe], which will emit a [List] of [AutoCompleteResult] of [Type] (or nothing)
+     * @return a [Maybe], which will emit an [ArrayList] of [AutoCompleteResult] of [Type] (or nothing)
```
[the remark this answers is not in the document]
HeinousGames: Mentioned in previous PR, can't visibly see the auto complete list remove/add items unless we use ArrayList.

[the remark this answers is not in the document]
wbaumann: We should be able to use the `toMutableList` piece or at least a `MutableList`. Where do we add/remove them?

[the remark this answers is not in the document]
HeinousGames: I reverted everything back to list implementation and the removals/additions to the autocomplete view list are working as originally intended. Not sure why it isn't giving me an issue now, but glad the list implementation will work, unless I missed one of the changes somewhere.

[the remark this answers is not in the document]
wbaumann: Niec work :)

app/src/main/java/co/smartreceipts/android/model/AutoCompleteMetadata.kt
```diff
@@ -0,0 +1,24 @@
+package co.smartreceipts.android.model
+
+import android.os.Parcelable
+import kotlinx.android.parcel.Parcelize
+
+@Parcelize
+data class AutoCompleteMetadata(
```
wbaumann: If you're feeling "fancy", you can convert this to a `sealed class`, so we can have separate fields for the two different types (Distance vs Receipts)

app/src/main/java/co/smartreceipts/android/distance/editor/DistanceCreateEditFragment.kt
```diff
@@ -343,21 +355,104 @@ class DistanceCreateEditFragment : WBFragment(), DistanceCreateEditView, View.On
...
+                            resultsAdapter.notifyDataSetChanged()
+                            Toast.makeText(context, R.string.result_restored, Toast.LENGTH_LONG).show()
+                        }
+                    }, {
+                        activity!!.runOnUiThread {
+                            Toast.makeText(activity, R.string.result_restore_failed, Toast.LENGTH_LONG).show()
+                        }
+                    })
+        } else {
+            presenter.updateDistanceCommentAutoCompleteVisibility(result.firstItem, false)
```
wbaumann: So in general here, I'd prefer the Fragment/Presenter/View flow to work as follows for consuming updates to the state:

1. Fragment calls onStart => Presenter calls subscribe => Presenter gets results and passes them to the View => The Fragment (which implements the View) then takes the result and applies it

When submitting state changes, it can work as follows:

1. The user performs some action => The fragment (aka view) emits an event in an Rx stream => The presenter listens to this event and does some work => The presenter informs the view that something changed via the View

The reason I mention this is that the current structure risks a memory leak if the processing gets interrupted (as the subscribe is not attached to a `CompositeDisposable`). 

So if you do this new structure, you'd have something like:

interface View {

fun onAutoCompleteVisibilityChange(): Observable<YourVisibilityChangeEvent>

}
```

Then the presenter would listen to this and perform the processing work whenever it gets an event

File: app/src/main/java/co/smartreceipts/android/model/AutoCompleteMetadata.kt (hunk 7 above)
Comment (HeinousGames) [reply app/src/main/java/co/smartreceipts/android/model/AutoCompleteMetadata.kt]: We also use this class for the Trip type, as they allow auto complete on the cost center. Although it appears I can't seal a class that is also a Parcel. This class needs to be a parcel because Distance, Receipt, and Trip are also Parcels.

Comment (wbaumann) [reply app/src/main/java/co/smartreceipts/android/model/AutoCompleteMetadata.kt]: So you can do some magic (I think) like:

```
sealed class AutoCompleteMetadata {
  abstract val nameField: Boolean
  // ...

  @Parcelize
  data class Receipt(
    override val nameField: Boolean,
    // ...
  ) : AutoCompleteMetadata()
}

HeinousGames: The AutoCompleteMetadata class has a constructor of four booleans, but you want me to remove the constructor and redo the class so they are abstract vals? Also, AutoCompleteMetadata is not an interface, but in your example you are treating it as such in the Receipt class. Interfaces are also not able to be sealed classes. I need a better example of how this is supposed to work.

app/src/main/java/co/smartreceipts/android/distance/editor/DistanceCreateEditFragment.kt
HeinousGames: First I'd like to point out that in the next pull request for PaymentMethodReimbursable, I have these presenter methods refactored to be more readable and take up less space. I would like to make this change in that PR, but I'm stumped as to what you want me to do.

I do notice that the IDE gives a warning in the fragments saying that the result of subscribe is unused (the memory leak you are referring to). What I can't seem to figure out is how do I do update this logic, yet still used the previously made method updateDistance from the DistanceCreateEditInteractor class? I can't find any examples in the app of the update being tracked outside of a completable (aside from calls to update that aren't returned at all).

I also am not sure if:
- The onAutoCompleteVisibilityChange method is Observing an Optional<Distance> object or some other object. I get the feeling that you want me to put the distance update logic in this method, but we already have it in the interactor.
- The view I'm adding this method to is the individual object creation view classes (distance, receipt, and trip), or just the AutoCompleteView class. It seems I have more control if I put the method in each object's respective view class instead of the AutoCompleteView class.

wbaumann: Could you point me to where you change it in the next MR?

The main thing I want to solve is that we have an untracked `.subscribe()`, which can lead to memory leaks. Here's how I'm thinking to go about this (although there may be better ways):

In your `AutoCompletePresenter` class, add the following line in your subscribe method:

```
fun subscribe() {
  // ...
  compositeDisposable.add(view.onAutoCompleteVisibilityChange()
         .flatMap { event -> interactor.performTheLogic(event) }
         .subscribe({
               view.showVisiblity(...)
         }))
}
```

This is a bit simplified, but that's the core idea. This does the following:

1. Rather than explicitly calling `             presenter.updateDistanceCommentAutoCompleteVisibility(result.firstItem, false)`, we emit an event each time instead. Then the presenter just listens to that event

From an event standpoint, check out some of the Views that have a `clicks()` type method, where it emits an event whenever a click happens. In our case, we can create a `PublishSubject` and just call `PublishSubject.onNext(YourEvent)` whenever we'd call this method. Does that make sense?

wbaumann: In terms of examples, check out how we use this `` and the `PublishSubject`s there:

* https://github.com/wbaumann/SmartReceiptsLibrary/blob/fea99cb4057cf82f5e605c7455cabc2b76a809c5/app/src/main/java/co/smartreceipts/android/distance/editor/DistanceCreateEditFragment.kt#L86

In the example above, whenever we click to save the distance, it emits a `Distance` event. Our `DistanceCreateEditView` has this as part of it's signature and the `DistanceCreateEditPresenter` will listen for these events and call to save it with our database

HeinousGames: @wbaumann 

In the next PR I updated the presenter methods to contain the autocomplete value and make it easier to create the new object to overwrite the old one. You can see an example of it here ( https://github.com/HeinousGames/SmartReceiptsLibrary/blob/6088cae833bceeacbaa28a4cb6896570c5a47cc0/app/src/main/java/co/smartreceipts/android/distance/editor/DistanceCreateEditPresenter.kt#L75). I also made these small changes to the ReceiptCreateEdit and TripCreateEdit presenters. This only made the code more readable and didn't fix the subscription memory leak you mention here. Given what you explained I was able to fix the memory leak warning. I made many changes to the fragments, presenters, views, etc. You can find all the changes in the PaymentMethodReimbursable commits on march 15 (https://github.com/HeinousGames/SmartReceiptsLibrary/commits/PaymentMethodReimbursable)

wbaumann: Gotcha. How do you want to proceed with these two PRs then? Is it best to merge this one in and then continue the review on the next one? I don't want to introduce too much back & forth for you

wbaumann: Or just close this one and move it all there?

HeinousGames: I think closing it should be fine. Anytime I made changes to this branch I merged it into the next one so it should have all reviewed changes there already.

HeinousGames:
After merging master to this branch, git has recognized there are no file changes. Closing.

PR #188, opened 2019-12-10
Pdf footer temp fix

HeinousGames:
report generation fails when footer text is blank. replacing with a space as a temp fix


wbaumann:
@HeinousGames - Seems like lots of conflicts/changes here outside of the commit name. Are you missing the latest `master` changes?

wbaumann:
Since we're going to cherry-pick this over to our current release, I'd really like to limit all possible changes here (i.e. no version updates, etc), so we can make this as small as possible to avoid conflicts. We can include the rest in a separate MR

HeinousGames:
> @HeinousGames - Seems like lots of conflicts/changes here outside of the commit name. Are you missing the latest `master` changes?

I made this PR as a direct branch from the release_4.20.2 branch, so it was missing commits from master. I merged and pushed the changes.

HeinousGames:
> Since we're going to cherry-pick this over to our current release, I'd really like to limit all possible changes here (i.e. no version updates, etc), so we can make this as small as possible to avoid conflicts. We can include the rest in a separate MR

The only file changed was DefaultPdfBoxPageDecorations, very minimal change, just not sure if it is the correct way to go, so could be temporary.

PR #191, opened 2019-12-16
[FLOSS] Removed Amazon dependency for floss flavor

JuliaSoboleva:
I started with the `OCR` function that uses `amazonaws` dependency. 

So, I created `aws` module which puts together all code that uses amazon dependency (all `cognito` and `s3` classes) and `core` module that contains the classes that all modules need (such as `Logger` for example). Each module has its own `build.gradle`, so now amazon dependencies are moved from main to `aws` module and further all not-floss dependencies will be moved to appropriate modules.

`CognitoManager` and `OcrManager` now have interfaces, real and `NoOp` implementations. There are instructions for Dagger in the `FlavorModule` class on which implementation should be provided for current flavor (`NoOp`s are provided just for `floss`).

It was an interesting part of removing `aws` dependencies from the apk. Dagger creates factories for all injectable classes even if they're not used (I mean real implementations of `CognitoManager` and `OcrManager` ), so I worked with our `proguard-rules` file. 
I added a new `flavor-not-floss-rules` file and transferred all rules for amazon from the main file to it. This new proguard file is used for all flavors except `floss`.

In this way, we're getting `flossFlavorRelease` apk without any `amazonaws` dependency. Proguard/R8 removes it for us.

Further, dependencies on `Google` and `Firebase` will also be separated in a similar way. As I understand, that's what F-Droid needs, right?

BTW, I found this rules in the `proguard-rules` file:

```
# Keep all of our classes as they are
-keep class co.smartreceipts.** { *; }
-dontwarn co.smartreceipts.**
-keep class wb.android.** { *; }
-dontwarn wb.android.**
-keep class wb.receipts.** { *; }
-dontwarn wb.receipts.**
-keep class wb.receiptspro.** { *; }
-dontwarn wb.receiptspro.**
```

Why does it exist? I commented out these lines for now because it keeps `amazonaws` dependency.


wbaumann:
Hi @JuliaSoboleva ,

Thanks for putting together this change. For your questions/notes:

> It was an interesting part of removing aws dependencies from the apk. Dagger creates factories for all injectable classes even if they're not used (I mean real implementations of CognitoManager and OcrManager ), so I worked with our proguard-rules file. I added a new flavor-not-floss-rules file and transferred all rules for amazon from the main file to it. This new proguard file is used for all flavors except floss.

Good find. I wouldn't have expected this :)

> Further, dependencies on Google and Firebase will also be separated in a similar way. As I understand, that's what F-Droid needs, right?

Yep. Exactly! I also think it has the side benefit of helping to enforce that we have a more modular architecture if we ever want to explore things like instant apps.

> BTW, I found this rules in the proguard-rules file. Why does it exist? I commented out these lines for now because it keeps amazonaws dependency.

I had originally added this to avoid the need to de-obfuscate our code when dealing with crash reports and crash logs. I think this is all automated now, but there's less risk in keeping our code public in that we're already open source :)

wbaumann:
For the GitHub trails, this change is one of the first to enable:

https://github.com/wbaumann/SmartReceiptsLibrary/issues/66

app/build.gradle
```diff
@@ -193,16 +199,11 @@ android {
...
-        implementation "com.amazonaws:aws-android-sdk-s3:$AWS_SDK_VERSION"
-
         // Dagger
-        implementation "com.google.dagger:dagger-android-support:$DAGGER_VERSION"
-        implementation "com.google.dagger:dagger:$DAGGER_VERSION"
         kapt "com.google.dagger:dagger-compiler:$DAGGER_VERSION"
         kapt "com.google.dagger:dagger-android-processor:$DAGGER_VERSION"
 
-        // Crashlytics
+        // Crashlyticsb
```
wbaumann: Typo

app/proguard-rules.pro
```diff
@@ -9,24 +9,25 @@
 
 ##### Our ProGuard configurations #####
 
+-whyareyoukeeping class com.amazonaws.*
```
wbaumann: Do we need the `whyareyoukeeping`? Can we use as a comment instead?

app/proguard-rules.pro
```diff
@@ -9,24 +9,25 @@
...
-# Keep all of our classes as they are
--keep class co.smartreceipts.** { *; }
--dontwarn co.smartreceipts.**
--keep class wb.android.** { *; }
--dontwarn wb.android.**
--keep class wb.receipts.** { *; }
--dontwarn wb.receipts.**
--keep class wb.receiptspro.** { *; }
--dontwarn wb.receiptspro.**
+## Keep all of our classes as they are
```
wbaumann: Delete? Or...?

aws/.gitignore
```diff
@@ -0,0 +1 @@
+/build
```
wbaumann: Let's use the default one:

https://github.com/github/gitignore/blob/master/Android.gitignore

Or simply update the one in the root to find the build directories recursively (to avoid the need for this when adding future modules)

JuliaSoboleva: I removed these generated gitignore files.
Our root gitignore works recursively as is, so it doesn't need to be updated as I see :)

wbaumann:
@JuliaSoboleva approved but the build failed. Not sure why

JuliaSoboleva:
@wbaumann I've had some trouble with rebase. Did it now, should work I think

wbaumann:
@JuliaSoboleva Seemed to fail again. Maybe an edge case issue? I won't merge anything else until you're able to get it fixed, so you won't have to worry about a stream of rebases :)

JuliaSoboleva:
@wbaumann Looks like I forgot to fix some tests... Will fix :)

PR #192, opened 2019-12-17
Payment method reimbursable

HeinousGames:
- extended from the same branches as the previous PRs because it also involved a db update
- please check unit tests updates for each branch as a few were failing and I made changes to make them pass, but not sure if they were appropriate changes


app/src/main/java/co/smartreceipts/android/receipts/editor/ReceiptCreateEditFragment.java
```diff
@@ -955,6 +958,62 @@ public void displayTitle(@NotNull String title) {
...
+                if (nameBox.isPopupShowing()) {
+                    presenter.updateReceiptNameAutoCompleteVisibility(firstReceipt, true)
+                            .subscribe(() ->
+                                    getActivity().runOnUiThread(() -> {
+                                        resultsAdapter.remove(selectedItem);
+                                        resultsAdapter.notifyDataSetChanged();
+                                        showUndoSnackbar(selectedItem, position, true);
+                                    }), throwableConsumer);
+                } else {
+                    presenter.updateReceiptCommentAutoCompleteVisibility(firstReceipt, true)
```
wbaumann: These should be added to the list of subscriptions (and preferablly all logic should be moved into the presenter, so it can be tested)

app/src/main/java/co/smartreceipts/android/receipts/editor/ReceiptCreateEditFragment.java
```diff
@@ -981,4 +1040,35 @@ private String getFlexString(int id) {
         return getFlexString(flex, id);
     }
 
+    private void showUndoSnackbar(AutoCompleteResult<Receipt> result, int position, boolean useNameBox) {
```
wbaumann: If possible, we should really move to include this in the `View` interface and then all the logic that triggers this can be in our `Presenter`, so we can test it

app/src/main/java/co/smartreceipts/android/trips/editor/TripCreateEditFragment.java
```diff
@@ -142,6 +145,9 @@
     // Misc Views
     private View focusedView;
 
+    private AutoCompleteArrayAdapter<Trip> resultsAdapter;
+    private Snackbar snackbar;
+    private boolean shouldHideResults;
```
wbaumann: What's this variable for?

app/src/main/java/co/smartreceipts/android/trips/editor/TripCreateEditFragment.java
```diff
@@ -461,19 +467,23 @@ private String getFlexString(int id) {
...
-        if (field == TripAutoCompleteField.Name) {
-            nameBox.setAdapter(resultsAdapter);
-            nameBox.showDropDown();
-        } else if (field == TripAutoCompleteField.Comment) {
-            commentBox.setAdapter(resultsAdapter);
-            commentBox.showDropDown();
-        } else if (field == TripAutoCompleteField.CostCenter) {
-            costCenterBox.setAdapter(resultsAdapter);
-            costCenterBox.showDropDown();
+    public void displayAutoCompleteResults(@NotNull AutoCompleteField field, @NotNull ArrayList<AutoCompleteResult<Trip>> autoCompleteResults) {
```
wbaumann: Can we make a presenter that encapsulates more of this? I'd like to share the logic across each

app/src/test/java/co/smartreceipts/android/persistence/database/tables/DistanceTableTest.java
```diff
@@ -232,14 +250,29 @@ public void onUpgradeFromV18() {
 
     }
 
+    @Test
+    public void onUpgradeFromV20() {
```
wbaumann: Looks good here!

app/src/main/java/co/smartreceipts/android/trips/editor/TripCreateEditFragment.java
HeinousGames: When the user picks an autocomplete result, the autocomplete view would still show what was selected as an option, because we altered the logic to show auto complete suggestions for infinite amount of characters in the view.

app/src/main/java/co/smartreceipts/android/trips/editor/TripCreateEditFragment.java
HeinousGames: The presenter class already has the displayAutoCompleteResults method. Do you want me to somehow move the logic called in it to the presenter class?

app/src/main/java/co/smartreceipts/android/receipts/editor/ReceiptCreateEditFragment.java
HeinousGames: Done!

HeinousGames:
> Let's stick with `listOf`. It's better practice to stick with the interface (i.e. `List`) instead of implementation `ArrayList` when possible

Just noticed that you asked about this. We weren't allowed to add/remove auto complete view objects from the adapter if we weren't using an array list. Had to refactor adapter, methods, and unit tests to use arraylist instead.

app/src/main/java/co/smartreceipts/android/autocomplete/AutoCompleteInteractor.kt
```diff
@@ -34,42 +40,65 @@ class AutoCompleteInteractor<Type> constructor(private val provider: AutoComplet
      * @param field the [AutoCompleteField] to use
      * @param input the current user input [CharSequence]
      *
-     * @return a [Maybe], which will emit a [List] of [AutoCompleteResult] of [Type] (or nothing)
+     * @return a [Maybe], which will emit an [ArrayList] of [AutoCompleteResult] of [Type] (or nothing)
      */
-    fun getAutoCompleteResults(field: AutoCompleteField, input: CharSequence) : Maybe<List<AutoCompleteResult<Type>>> {
+    fun getAutoCompleteResults(field: AutoCompleteField, input: CharSequence) : Maybe<ArrayList<AutoCompleteResult<Type>>> {
```
wbaumann: Let's change this to `List` instead of `ArrayList`:

https://stackoverflow.com/questions/10328658/should-i-return-list-or-arraylist

app/src/main/java/co/smartreceipts/android/autocomplete/AutoCompleteView.kt
```diff
@@ -26,6 +26,11 @@ interface AutoCompleteView<Type> {
      * @param results a [List] of [AutoCompleteResult] of [Type]
      */
     @UiThread
-    fun displayAutoCompleteResults(field: AutoCompleteField, results: List<AutoCompleteResult<Type>>)
+    fun displayAutoCompleteResults(field: AutoCompleteField, results: ArrayList<AutoCompleteResult<Type>>)
```
wbaumann: Same here to `List<>`

app/src/main/java/co/smartreceipts/android/autocomplete/AutoCompleteInteractor.kt
```diff
@@ -34,42 +40,65 @@ class AutoCompleteInteractor<Type> constructor(private val provider: AutoComplet
...
+                                        else -> false
+                                    }
+
+                                    if (!removedByUser) {
+                                        val displayName = resultsChecker.getValue(field, it)
+                                        // Only allow input with new display names
+                                        if (!resultsSet.contains(displayName)) {
+                                            val result = AutoCompleteResult(displayName, it)
+                                            resultsSet[displayName] = result
+                                            results.add(result)
```
wbaumann: I assume you used `ArrayList` for the `.add` function. In Kotlin, it's more idiomatic to do: `mutableListOf<>()` for this type of stuff.

You can also use the `+` operator to combine two lists like:

```
val combinedList = list1 + list2
```

app/src/main/java/co/smartreceipts/android/trips/editor/TripCreateEditFragment.java
```diff
@@ -263,13 +285,30 @@ public void onResume() {
         if (focusedView != null) {
             focusedView.requestFocus(); // Make sure we're focused on the right view
         }
+
+        compositeDisposable = new CompositeDisposable();
+        compositeDisposable.add(getHideAutoCompleteVisibilityClick()
```
wbaumann: This looks good! Let's now move it into a Presenter class like `AutoCompeltePresenter` or `TripAutoCompeltePresenter`, so we can isolate all the logic in that one class (instead of spreading it across the fragment)

app/src/main/java/co/smartreceipts/android/trips/editor/TripCreateEditFragment.java
```diff
@@ -533,4 +576,117 @@ public void hideTooltip() {
...
+            });
+        } else {
+            getActivity().runOnUiThread(() ->
+                    Toast.makeText(getContext(), R.string.result_restore_failed, Toast.LENGTH_LONG).show());
+        }
+    }
+
+    @NotNull
+    @Override
+    public Observable<Pair<Trip, Trip>> getHideAutoCompleteVisibilityClick() {
```
wbaumann: Let's change the signature here from `<Pair<Trip, Trip>` to a new dataclass like:

```
data class AutoCompleteClickEvent<Type> (
  val from: Type,
  val to: Type
)
```

So it's a bit clearer what the `first` and `second` in this pair do

app/src/main/java/co/smartreceipts/android/persistence/database/tables/PaymentMethodsTable.java
```diff
@@ -90,6 +92,12 @@ public synchronized void onUpgrade(@NonNull SQLiteDatabase db, int oldVersion, i
         if (oldVersion <= 18) { // v18 => 19 added UUID column
             onUpgradeToAddUUID(db, oldVersion);
         }
+
+        if (oldVersion <= 20) { // v19 => 20 added reimbursable column
```
wbaumann: @HeinousGames - Let's change it to version `21`, since we didn't cut this into this new release (which is for v20)

app/src/main/java/co/smartreceipts/android/distance/editor/DistanceCreateEditPresenter.kt
```diff
@@ -48,6 +48,24 @@ class DistanceCreateEditPresenter @Inject constructor(
             .distinctUntilChanged()
             .doOnNext { interactor.deleteDistance(it) }
             .subscribe { view.present(UiIndicator.success()) })
+
+        compositeDisposable.add(view.hideAutoCompleteVisibilityClick
```
wbaumann: Looks great! Now as a next step, let's pull this into a separate presenter class entirely like:

```
class DistanceAutoCompletePresenter{
 ...
}
```

The idea here is that we can isolate auto-complete logic from create/edit logic from "other" logic. So we can easily add/update/delete a single piece of business logic where we need to.

This idea is heavily inspired by the concept of "Clean Architecture". Take a look: https://rubygarage.org/blog/clean-android-architecture

app/src/main/java/co/smartreceipts/android/autocomplete/AutoCompleteInteractor.kt
HeinousGames: This needs to be an array list, otherwise the Receipt and Trip editing screens will crash when trying to remove an item from the drop down list. Lists are not allowed to be modified, but array lists are. A list will work on the Distance editing screen because it is kotlin class, but not in Trip or Receipt because they are java classes.

app/src/main/java/co/smartreceipts/android/autocomplete/AutoCompleteInteractor.kt
HeinousGames: In this instance I'm checking each item individually to see if they need to be added to the result list. I don't need to merge two lists, is it redundant to do
`results = results + result` instead of `results.add(result)`?

HeinousGames:
I seem to have found an issue. Each auto complete value is tied to an object (distance, trip, or receipt). If there are two receipts with the same name or comment, then that value can be shown in auto complete multiple times. We have logic to hide duplicates from appearing in the auto complete list, but after you delete one and type similar text, one of the other ones will take its place. Do you have any thoughts on how we can fix that, or do you think there are a low amount of users who name receipts and comments the same?

app/src/main/java/co/smartreceipts/android/trips/editor/TripCreateEditFragment.java
HeinousGames: Done!

app/src/main/java/co/smartreceipts/android/persistence/database/tables/PaymentMethodsTable.java
HeinousGames: Done!

app/src/main/java/co/smartreceipts/android/distance/editor/DistanceCreateEditPresenter.kt
HeinousGames: Done!

app/src/main/java/co/smartreceipts/android/trips/editor/TripCreateEditFragment.java
HeinousGames: Done!

app/src/main/java/co/smartreceipts/android/autocomplete/AutoCompleteInteractor.kt
wbaumann: Let's use a `MutableList` then. It's a Kotlin interface abstraction on top of that

wbaumann:
For this:

> Do you have any thoughts on how we can fix that, or do you think there are a low amount of users who name receipts and comments the same?

Great point. Can we mark all items with that name/comment as hidden when you user selects to hide it? Rather than just hiding that single one?

app/src/main/java/co/smartreceipts/android/autocomplete/AutoCompleteInteractor.kt
wbaumann: My reasoning for the `List` is that we can do this:

1. Always pass a `List` as it's immutable
2. The user clicks to "hide" an entry
3. We perform some logic to do this
4. We then return a new `List` with the updated set of results

app/src/main/java/co/smartreceipts/android/autocomplete/AutoCompleteView.kt
```diff
@@ -26,6 +27,11 @@ interface AutoCompleteView<Type> {
...
 
+    val hideAutoCompleteVisibilityClick: Observable<AutoCompleteClickEvent<Type>>
+    val unHideAutoCompleteVisibilityClick: Observable<AutoCompleteClickEvent<Type>>
+
+    fun hideAutoCompleteValue(wasSuccessfullyHidden: Boolean)
```
wbaumann: What does `wasSuccessfullyHidden` mean here?

wbaumann: If this is to handle errors, can we add a new method instead like:

```
fun displayAutoCompleteError()
```

I think that approach is a bit more readable

app/src/main/java/co/smartreceipts/android/distance/editor/DistanceCreateEditFragment.kt
```diff
@@ -341,23 +364,117 @@ class DistanceCreateEditFragment : WBFragment(), DistanceCreateEditView, View.On
...
-                DistanceAutoCompleteField.Comment -> {
-                    text_distance_comment.setAdapter(resultsAdapter)
-                    text_distance_comment.showDropDown()
+            } else {
+                shouldHideResults = false
+            }
+        }
+    }
+
+    override fun onClick(removeAutoCompleteResult: Boolean, position: Int) {
```
wbaumann: Does this method come from `AutoCompleteArrayAdapter.ClickListener`? I'd love to pull more of this logic into the Presenter if possible, so we can test more of it

wbaumann: The main idea here is that I want to push for our Fragments to be as "dumb" as possible, so all they know is about "display this" or "the user clicked that"

app/src/main/java/co/smartreceipts/android/distance/editor/DistanceCreateEditFragment.kt
```diff
@@ -341,23 +364,117 @@ class DistanceCreateEditFragment : WBFragment(), DistanceCreateEditView, View.On
...
+                    true -> {
+                        autoCompleteType = AutoCompleteType.Location
+                        newDistance = DistanceBuilderFactory(oldDistance)
+                                .setLocationHiddenFromAutoComplete(true)
+                                .build()
+                    }
+                    false -> {
+                        autoCompleteType = AutoCompleteType.Comment
+                        newDistance = DistanceBuilderFactory(oldDistance)
+                                .setCommentHiddenFromAutoComplete(true)
```
wbaumann: So for a more concrete example of moving this logic to the presenter, let's say we instead emit a new event that just contains the following:

```
data class AutoCompleteUpdateEvent<Distance>(
  val item: Distance,
  val type: AutoCompleteType
)
```

Then our presenter can do something like:

```
when {
  Distance.Location -> DistanceBuilderFactory(event.item).setLocationHiddenFromAutoComplete(true).build()
  Distance.Comment -> DistanceBuilderFactory(event.item).setCommentHiddenFromAutoComplete(true).build()
}

Comment (wbaumann) [reply app/src/main/java/co/smartreceipts/android/distance/editor/DistanceCreateEditFragment.kt]: This will allow us to test all those cases in our presenter and pull more logic outside of our fragment

File: app/src/main/java/co/smartreceipts/android/model/AutoCompleteClickEvent.kt [hunk 16]
```diff
@@ -0,0 +1,6 @@
+package co.smartreceipts.android.model
+
+data class AutoCompleteClickEvent<Type>(
```
wbaumann: Check my comment above here. I'd like to change this signature to:

```
data class AutoCompleteUpdateEvent<Type>(
  val item: Type,
  val type: AutoCompleteType
)
```

app/src/main/java/co/smartreceipts/android/receipts/editor/ReceiptCreateEditFragment.java
```diff
@@ -968,4 +1040,66 @@ private String getFlexString(int id) {
...
+    private void showUndoSnackbar(AutoCompleteResult<Receipt> result, @NotNull AutoCompleteType autoCompleteType) {
+        View view = getActivity().findViewById(R.id.update_receipt_layout);
+        snackbar = Snackbar.make(view, getString(
+                R.string.item_removed_from_auto_complete, result.getDisplayName()), Snackbar.LENGTH_LONG);
+        snackbar.setAction(R.string.undo, v -> {
+            Receipt newReceipt;
+            final Receipt oldReceipt = result.getFirstItem();
+            if (autoCompleteType == AutoCompleteType.Name) {
+                newReceipt = new ReceiptBuilderFactory(oldReceipt)
+                        .setNameHiddenFromAutoComplete(false)
```
wbaumann: Then similarly, we can pull this logic into the ReceiptAutoComplete presenter

app/src/main/java/co/smartreceipts/android/distance/editor/DistanceCreateEditFragment.kt
HeinousGames: Instead of the click listener I pass the autocomplete view to the adapter. The view then calls some methods that I separated the onClick method into. Is there a way I can make the fragment more dumb than this?

HeinousGames:
> For this:
> 
> > Do you have any thoughts on how we can fix that, or do you think there are a low amount of users who name receipts and comments the same?
> 
> Great point. Can we mark all items with that name/comment as hidden when you user selects to hide it? Rather than just hiding that single one?

Not sure. I suspect that we could bring the table controllers into the edit fragment classes to get a reference to all of the other objects in the table. Then when a user deletes one from autocomplete we can go through the list of objects and find any that have the same value and hide those too. Implementing the table events listener for the edit fragment classes seemed like a lot just for this though. Thoughts?

wbaumann:
For:

> Not sure. I suspect that we could bring the table controllers into the edit fragment classes to get a reference to all of the other objects in the table. Then when a user deletes one from autocomplete we can go through the list of objects and find any that have the same value and hide those too. Implementing the table events listener for the edit fragment classes seemed like a lot just for this though. Thoughts?

I agree. Let's just start by leaving this bug in place, and then we'll work on fixing that in an upcoming MR. This will at least give us a first iteration to get it working properly. I have a few ideas for how to properly address this in the future, but I think one way is to decouple the "autocomplete" suggestions from the actual data. Let's discuss more once we've merged this all in :)

app/src/main/java/co/smartreceipts/android/autocomplete/AutoCompleteView.kt
```diff
@@ -26,6 +27,14 @@ interface AutoCompleteView<Type> {
      * @param results a [List] of [AutoCompleteResult] of [Type]
      */
     @UiThread
-    fun displayAutoCompleteResults(field: AutoCompleteField, results: List<AutoCompleteResult<Type>>)
+    fun displayAutoCompleteResults(field: AutoCompleteField, results: MutableList<AutoCompleteResult<Type>>)
 
+    val hideAutoCompleteVisibilityClick: Observable<AutoCompleteUpdateEvent<Type>>
+    val unHideAutoCompleteVisibilityClick: Observable<AutoCompleteUpdateEvent<Type>>
+
+    fun fillValueField(position: Int)
```
wbaumann: Can we pass the object here instead of the `position`? While `position` works fine, it's a little brittle IMO, so it's less safe that the object.

app/src/main/java/co/smartreceipts/android/autocomplete/AutoCompleteView.kt
```diff
@@ -26,6 +27,14 @@ interface AutoCompleteView<Type> {
...
+    fun deleteAutoCompleteValueFromDB(position: Int)
```
wbaumann: Does this method overlap with the `hideAutoCompleteVisibilityClick` one? I'd prefer to keep the higher level logic functions outside of the Views.

Here's the ideal structure:

1. `Fragment` => Supports only "display this" or "that data was clicked"
2. `Presenter` => Performs the logic to assemble what to display or route the click events to the appropriate places (e.g. Interactors for business logic)
3. `Interactor` => Performs the action business logic like deleting it from the DB

app/src/main/java/co/smartreceipts/android/distance/editor/DistanceAutoCompletePresenter.kt
```diff
@@ -0,0 +1,58 @@
...
+        compositeDisposable.add(view.hideAutoCompleteVisibilityClick
+                .flatMap { t ->
+                    when (t.type) {
+                        DistanceAutoCompleteField.Location -> interactor.updateDistance(t.item, DistanceBuilderFactory(t.item)
+                                .setLocationHiddenFromAutoComplete(true)
+                                .build())
+                        DistanceAutoCompleteField.Comment -> interactor.updateDistance(t.item, DistanceBuilderFactory(t.item)
+                                .setCommentHiddenFromAutoComplete(true)
+                                .build())
+                        else -> interactor.updateDistance(t.item, t.item)
```
wbaumann: I think we can do `throw UnsupportedOperationException(...)` here, since we only support these two types, correct?

app/src/main/java/co/smartreceipts/android/distance/editor/DistanceCreateEditFragment.kt
wbaumann: You handled it below, but the basic structure of this all comes from the following:

1. The more dumb the Fragment is, the easier it is to test
2. Fragments are pretty much impossible to test by default in Android :)

Because of this, I try to pull "all" logic aside from "display X" or "Y was clicked" outside of the fragment. Then you can test all those flows in your Presenters.

This is a better example that my poor comment above:

https://github.com/wbaumann/SmartReceiptsLibrary/pull/192#discussion_r399722983

app/src/main/java/co/smartreceipts/android/distance/editor/DistanceCreateEditFragment.kt
```diff
@@ -341,23 +367,90 @@ class DistanceCreateEditFragment : WBFragment(), DistanceCreateEditView, View.On
...
+            } else {
+                text_distance_comment.setText(autoCompleteVisibilityItem!!.displayName)
+                text_distance_comment.setSelection(text_distance_comment.text.length)
+                text_distance_comment.dismissDropDown()
             }
+            SoftKeyboardManager.hideKeyboard(focusedView)
+        }
+    }
+
+    override fun deleteAutoCompleteValueFromDB(position: Int) {
```
wbaumann: This may just be a naming thing, but this is the one method that I think we can clean up a bit. Ideally, the fragment has no idea that there's even a database behind it. All it knows is that we're trying to "hide item".

For this method, what are we trying to perform? Emit an event to show it was hidden?

wbaumann:
Btw - nice work so far! This is really coming along well. Let me know if anything is unclear or doesn't make enough sense as I keep providing feedback. I'm trying to spend a bit more time on this one, so the next set of changes will be more immediately clear as to the goals/benefits.

app/src/main/java/co/smartreceipts/android/distance/editor/DistanceAutoCompletePresenter.kt
wbaumann: Also to double check. Do we need the else block? Or can we delete it there?

app/src/main/java/co/smartreceipts/android/distance/editor/DistanceCreateEditFragment.kt
```diff
@@ -341,23 +367,82 @@ class DistanceCreateEditFragment : WBFragment(), DistanceCreateEditView, View.On
...
+        }
+        SoftKeyboardManager.hideKeyboard(focusedView)
+    }
+
+    override fun hideAutoCompleteClick(autoCompleteResult: AutoCompleteResult<Distance>) {
+        SoftKeyboardManager.hideKeyboard(focusedView)
+        autoCompleteVisibilityItem = autoCompleteResult
+        positionToUpdateVisibility = resultsAdapter.getPosition(autoCompleteResult)
+        val oldDistance = autoCompleteResult.firstItem
+        when(text_distance_location.isPopupShowing) {
```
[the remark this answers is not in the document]
wbaumann: Btw - does this approach all make sense in terms of how/why we're structuring it this way?

app/src/main/java/co/smartreceipts/android/distance/editor/DistanceAutoCompletePresenter.kt
```diff
@@ -0,0 +1,58 @@
...
+        compositeDisposable.add(view.hideAutoCompleteVisibilityClick
+                .flatMap { t ->
+                    when (t.type) {
+                        DistanceAutoCompleteField.Location -> interactor.updateDistance(t.item, DistanceBuilderFactory(t.item)
+                                .setLocationHiddenFromAutoComplete(true)
+                                .build())
+                        DistanceAutoCompleteField.Comment -> interactor.updateDistance(t.item, DistanceBuilderFactory(t.item)
+                                .setCommentHiddenFromAutoComplete(true)
+                                .build())
+                        else -> throw UnsupportedOperationException()
```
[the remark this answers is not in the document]
HeinousGames: Since we're passing AutoCompleteField the when statement is expecting options for trips and receipts.

app/src/main/java/co/smartreceipts/android/autocomplete/AutoCompleteView.kt
```diff
@@ -26,6 +27,14 @@ interface AutoCompleteView<Type> {
...
-    fun displayAutoCompleteResults(field: AutoCompleteField, results: List<AutoCompleteResult<Type>>)
+    fun displayAutoCompleteResults(field: AutoCompleteField, results: MutableList<AutoCompleteResult<Type>>)
 
+    val hideAutoCompleteVisibilityClick: Observable<AutoCompleteUpdateEvent<Type>>
+    val unHideAutoCompleteVisibilityClick: Observable<AutoCompleteUpdateEvent<Type>>
+
+    fun fillValueField(autoCompleteResult: AutoCompleteResult<Type>)
+    fun hideAutoCompleteClick(autoCompleteResult: AutoCompleteResult<Type>)
+    fun removeValueFromDropDown()
+    fun unHideAutoCompleteClick()
```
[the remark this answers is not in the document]
HeinousGames: The hideClick method sends the hide event to make the interactor change the db value. If it succeeds the view calls removeValueFromDropDown so the user sees it worked. I have renamed these methods.

app/src/main/java/co/smartreceipts/android/distance/editor/DistanceCreateEditFragment.kt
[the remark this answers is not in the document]
HeinousGames: You mentioned that the main reason is for unit testing? Otherwise we can't mock our activities/classes very well? By having them in presenters/interactors the testing is more fluid?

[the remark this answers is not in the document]
wbaumann: Correct (although there are some architecture benefits too). From a testing standpoint, it's because of the following:

* The Android system is very flake-y and hard to reliability test. For instance, in order to test a fragment, you need to be able to "drive" the entire lifecycle from `onCreate()` to `onDestroy()`. Robolectric tries and is the best, but it's a hard problem. Different OS levels have different configurations and just make it more difficult. (Also - you can't really mock any parts of the core Android system).
* Even if you can emulate the Android system (e.g. with Robolectric), it's much slower to spin this up. You're much better off testing pure Java where you can (for speed) and isolating the Android components to small portions for integration testing (via Espresso).

Our approach is based off the [iOS VIPER Pattern](https://www.objc.io/issues/13-architecture/viper/). We actually use a hybrid that has a variety of MVVM approaches (and if you are starting a new app today, I would 100% just use MVVM as it's the new standard by Google). It has the following in terms of structure in an ideal app (note: we don't follow all of this perfectly :) ):

* `Presenter`. Pure Java with no Android components. It controls the view logic and routes view clicks
* `Interactor`. Pure Java and no Android components. It handles all business logic
* `Fragment`. Only communicates with the presenter and is responsible for all things "Android"

With this approach, you can mock the `View` that you pass into each `Presenter` and test all potential flows. This has a few benefits:

1. You can be confident that if everything is wired up properly, it all works correctly
2. From an architecture standpoint, you can easily share the same logic across multiple fragments (like we're doing here). More code reusability means less bugs and more reliability

BUT it does have one major con: you don't actually test that the Fragment is correctly wired up with everything. This is where Espresso comes in, since it allows you to test that the entire flow works (at least for core flows). Espresso is much more flake-y than the rest of our tests (which should always work), but it'll help us keep confidence that we didn't break anything substantially.

Makes sense?

app/src/main/java/co/smartreceipts/android/autocomplete/AutoCompleteInteractor.kt
```diff
@@ -48,28 +54,51 @@ class AutoCompleteInteractor<Type> constructor(private val provider: AutoComplet
...
-                                    val displayName = resultsChecker.getValue(field, it)
-                                    // Only allow input with new display names
-                                    if (!resultsSet.contains(displayName)) {
-                                        val result = AutoCompleteResult(displayName, it)
-                                        resultsSet[displayName] = result
-                                        results.add(result)
-                                    } else {
-                                        resultsSet[displayName]!!.additionalItems.add(it)
+                                    // make sure value wasn't removed by user
+                                    val removedByUser = when (it) {
```
wbaumann, line 58: As a final step here, let's create a few tests to verify that these flows both:

1. Work
2. And don't accidentally get broken in the future :)

I think we'll want to do the following ones:

1. Test each of our new presenters (e.g. when we emit an `onHideAutocomplete(...)` event, do we get a specific event in response. You should be able to follow the approach in the other tests
2. For the `Interactor`, let's test that the business logic works. For example when it's a `Receipt.Name` field, the result should update that portion of the receipt

Let me know if that all makes sense!

HeinousGames: I just committed some changes for testing the presenters, but I'm having trouble with the interactors, and this is leading me to believe we haven't made our fragments fully dumb yet.

I'm not sure if you noticed this, but in all of our fragments I have setup three variables; an int to track position, an auto complete field to track the field, and a visibility item to know which item to remove from the drop down. We don't pass enough data from the fragments to the presenters for us to know which items to remove (and re-add if the user undoes the removal) so I end up storing it in these variables. You can see them at this line in the DistanceCreateEditFragment https://github.com/wbaumann/SmartReceiptsLibrary/blob/e23cc48fa9192b9f3e71a78fc1e4de8dbb4a5811/app/src/main/java/co/smartreceipts/android/distance/editor/DistanceCreateEditFragment.kt#L96
I realized that my presenter tests do test if the view's methods get called, but if anybody were to remove or modify these vars in the fragments, the method logic in the fragment class would still break. So the unit tests are only testing that the data is making it to the presenter and that the presenter is attempting to make the fragment do stuff, but not that any of it was successful.

Regarding tests for the interactor. When we send an auto complete update event to the presenter to update the object, we never actually check the object has the correct setting before removing it from the drop down. We just assume that as long as the update method spit out an object, that the object is correct. We don't actually do anything with the object, so it is very difficult for me to test that an object spit out has the correct settings flagged when we don't keep reference to it. Is there a way to test our interactor code that I'm not seeing?

wbaumann: Makes sense about this:

> So the unit tests are only testing that the data is making it to the presenter and that the presenter is attempting to make the fragment do stuff, but not that any of it was successful.

This is one of the boundaries in which it becomes difficult, since you need to manage some state (and Fragments are one place to save this). One approach to handling this is to take advantage of the `onSaveInstanceState` and immediately passing these variables to the Presenters to save. IIRC, you can check out the "Exchange Rate Presenter" to see an example of this.

For this:

> We don't actually do anything with the object, so it is very difficult for me to test that an object spit out has the correct settings flagged when we don't keep reference to it. Is there a way to test our interactor code that I'm not seeing?

That's fine then. It's really only handy when we have logic on top of whatever the database does

HeinousGames: > This is one of the boundaries in which it becomes difficult, since you need to manage some state (and Fragments are one place to save this). One approach to handling this is to take advantage of the `onSaveInstanceState` and immediately passing these variables to the Presenters to save. IIRC, you can check out the "Exchange Rate Presenter" to see an example of this.

Will, are you saying that I can save the position, auto complete field, and result object in the savedInstanceState so the presenters and views can interact with those values instead of the ones I previously created? I can make those changes if it's more appropriate.

wbaumann: Yea exactly. In an "ideal" situation, you should be able to rotate before saving and maintain the same state as before (note: the new `ViewModel` approach handles this automatically for us, but it's quite complex to re-write everything to this new way)

HeinousGames: @wbaumann I checked out a few presenters that had savedInstanceState passed to them, but still couldn't understand how to utilize it for our situation. Nonetheless, I did so some refactoring and moved more logic to the presenter, please check it out and let me know if it is okay or not. I also added some code to fix an issue where a user could undo a remove and accidentally add the field into the edit text view.

wbaumann: Cool that works. Thanks for the updates!

PR #199, opened 2020-01-22
[FLOSS] Removed Firebase dependency for floss flavor

JuliaSoboleva:
Firebase dependencies are used for push messages and analytics. 

All analytics and logging code was separated into a new `analytics` module.

The thing is, the Firebase contains their proguard rules inside. So to get rid of their code in the final floss apk file I had to add special `flavorDimensions "isFloss"` to `analytics` and `push` modules to be able to not include Firebase dependency to `FlossFlavor`.

For this purpose, I added `missingDimensionStrategy 'isFloss', 'notFloss'` for all flavors and `missingDimensionStrategy 'isFloss', 'floss'` for `FlossFlavor` in the `build.gradle(app)` `productFlavors` section.

We also have modules that don't need `floss`/`notFloss` flavors (`aws` and `core`) but since all modules depend on the `analytics` module the project can't be configured without `missingDimensionStrategy` in these `aws` and `core` `defaultConfig` section. So I specified `notFloss` as default, but I'm sure that in the end, it is overridden by flavors specified at `productFlavors` section in the `build.gradle(app)`. I need to mention it because Android Studio shows this as a conflict when the `FlossFlavor` build variant is selected but it actually works without problems.

There was also one strange [issue with `Dagger`](https://github.com/google/dagger/issues/970) because of which we need to use transitive dependency for  `com.google.firebase:firebase-messaging` at the `push` module. I left links related to this problem in the `build.gradle(push)` code.

The next step is removing Google Drive dependency from `floss` I think :)


wbaumann:
Sounds great and thank you for explaining this item. Nice work!

> We also have modules that don't need floss/notFloss flavors (aws and core) but since all modules depend on the analytics module the project can't be configured without missingDimensionStrategy in these aws and core defaultConfig section. So I specified notFloss as default, but I'm sure that in the end, it is overridden by flavors specified at productFlavors section in the build.gradle(app). I need to mention it because Android Studio shows this as a conflict when the FlossFlavor build variant is selected but it actually works without problems.

wbaumann:
@JuliaSoboleva - My fault - I thought I had merged this in, but I must not have... Would you be able to rebase it in with some of the newer changes? Apologies for the extra work here

JuliaSoboleva:
Sure, I'll take a look on Monday :)

пн, 27 янв. 2020 г., 0:10 Will Baumann <notifications@github.com>:

wbaumann:
Actually ignore that. The Gitlab cache had an issue refreshing that

PR #201, opened 2020-01-26
Import despite missing files

HeinousGames:
Allows for up to 5 missing files, but when complete still says "imported all files successfully." Would like to say how many files are missing instead, or the actual names of the receipts that have images that are missing. Not sure how to keep track of that info outside of sharedpreferences. After your review/approval, I'd like to send a version with this code to our two users who are having issues with this specifically.


app/src/main/java/co/smartreceipts/android/sync/drive/DriveServiceHelper.java
```diff
@@ -144,27 +147,26 @@ public Completable deleteFile(String fileId) {
...
+        int read;
+        while ((read = is.read(buffer)) != -1) {
+          fileOutputStream.write(buffer, 0, read);
+        }
+        return Single.just(Optional.of(downloadLocationFile));
+      } catch (FileNotFoundException e) {
+        Logger.error(DriveServiceHelper.this, "Failed write file with FileNotFoundException: ", e);
+        return Single.error(e);
+      }
+    } catch (GoogleJsonResponseException e) {
```
wbaumann: Just to confirm - this is the type of error that we're planning to ignore, right? If so, can we update the `Logger` in here to use a logger.warn statement instead?

app/src/main/java/co/smartreceipts/android/sync/drive/managers/DriveRestoreDataManager.java
```diff
@@ -80,21 +81,24 @@ public DriveRestoreDataManager(@NonNull Context context,
...
         Preconditions.checkNotNull(downloadLocation);
 
         return mDriveStreamsManager.getFilesInFolder(remoteBackupMetadata.getId().getId())
                 .flatMap(fileList -> {
-                    List<java.io.File> javaFileList = new ArrayList<>();
+                    List<Optional<java.io.File>> javaFileList = new ArrayList<>();
                     for (File file : fileList.getFiles()) {
                         String filename = file.getId() + "__" + file.getOriginalFilename();
-                        javaFileList.add(mDriveStreamsManager.download(file.getId(), new java.io.File(downloadLocation, filename)).blockingGet());
+                        Optional<java.io.File> single = mDriveStreamsManager.download(file.getId(), new java.io.File(downloadLocation, filename)).blockingGet();
```
wbaumann: Let's name this `file` or `optionalFile` instead of `single` (to separate it from the Rx `Single`)

app/src/main/java/co/smartreceipts/android/sync/drive/managers/DriveRestoreDataManager.java
```diff
@@ -116,19 +120,24 @@ public DriveRestoreDataManager(@NonNull Context context,
                         if (f.getModifiedTime().getValue() > date.getTime()) {
                             String filename = ModelUtils.getFormattedDate(
                                     new Date(f.getModifiedTime().getValue()), TimeZone.getDefault(), mContext, "-") + "_" + f.getOriginalFilename();
-                            javaFileList.add(mDriveStreamsManager.download(f.getId(), new java.io.File(downloadLocation, filename)).blockingGet());
+                            Optional<java.io.File> single = mDriveStreamsManager.download(f.getId(), new java.io.File(downloadLocation, filename)).blockingGet();
```
wbaumann: Same here

app/src/main/java/co/smartreceipts/android/sync/drive/managers/DriveRestoreDataManager.java
```diff
@@ -164,7 +173,16 @@ public DriveRestoreDataManager(@NonNull Context context,
                 })
                 .flatMapSingle(partialReceipt -> {
                     Logger.debug(DriveRestoreDataManager.this, "Downloading file for partial receipt: {}", partialReceipt.driveId);
-                    return downloadFileForReceipt(partialReceipt, downloadLocation);
+                    Single<Optional<Optional<java.io.File>>> singleOptionalOptional = downloadFileForReceipt(partialReceipt, downloadLocation);
+                    Optional<Optional<java.io.File>> optionalOptional = singleOptionalOptional.blockingGet();
```
wbaumann: How come this is an `Optional<Optional<>>`? I think it should just be an `Optional<>`

app/src/main/java/co/smartreceipts/android/sync/drive/managers/DriveRestoreDataManager.java
```diff
@@ -261,7 +279,7 @@ public DriveRestoreDataManager(@NonNull Context context,
         });
     }
 
-    private Single<Optional<java.io.File>> downloadFileForReceipt(@NonNull final PartialReceipt partialReceipt, @NonNull final java.io.File inDirectory) {
+    private Single<Optional<Optional<java.io.File>>> downloadFileForReceipt(@NonNull final PartialReceipt partialReceipt, @NonNull final java.io.File inDirectory) {
```
wbaumann: I don't think we need the second optional wrapper here. There's no real need for it twice

app/src/main/java/co/smartreceipts/android/sync/drive/managers/DriveRestoreDataManager.java
```diff
@@ -164,7 +173,15 @@ public DriveRestoreDataManager(@NonNull Context context,
                 })
                 .flatMapSingle(partialReceipt -> {
                     Logger.debug(DriveRestoreDataManager.this, "Downloading file for partial receipt: {}", partialReceipt.driveId);
-                    return downloadFileForReceipt(partialReceipt, downloadLocation);
+                    Single<Optional<java.io.File>> singleOptional = downloadFileForReceipt(partialReceipt, downloadLocation);
+                    Optional<java.io.File> optionalFile = singleOptional.blockingGet();
+                    if (!optionalFile.isPresent()) {
+                        missingFileCount.getAndIncrement();
+                        if (missingFileCount.get() > 5) {
```
wbaumann: Let's move this `5` into a `ALLOWED_DOWNLOAD_FAILURES` static int like:

```
private static final Integer ALLOWED_DOWNLOAD_FAILURES = 5;
```

Like this: https://stackoverflow.com/questions/47882/what-is-a-magic-number-and-why-is-it-bad

app/src/main/java/co/smartreceipts/android/sync/drive/DriveServiceHelper.java
HeinousGames: I updated the error to a warn, but for the GoogleJsonResponseException. The FileNotFoundException is a literal if the java file can't be created to put it in, but the GoogleException is the one we use to detect if we are avoiding the missing file abortion.

app/src/main/java/co/smartreceipts/android/sync/widget/backups/ImportRemoteBackupWorkerProgressDialogFragment.java
```diff
@@ -111,7 +111,11 @@ public void onResume() {
                     }
                 }, throwable -> {
                     analytics.record(new ErrorEvent(ImportRemoteBackupWorkerProgressDialogFragment.this, throwable));
-                    Toast.makeText(getActivity(), getString(R.string.IMPORT_ERROR), Toast.LENGTH_LONG).show();
+                    if (throwable.getMessage().contains("More than five missing files")) {
```
wbaumann: For this, let's change it to actually check the exception type (and we can return a different exception for these two cases). This version won't be as sustainable (if we ever change the messaging)

app/src/main/java/co/smartreceipts/android/sync/drive/managers/DriveRestoreDataManager.java
```diff
@@ -164,7 +175,15 @@ public DriveRestoreDataManager(@NonNull Context context,
                 })
                 .flatMapSingle(partialReceipt -> {
                     Logger.debug(DriveRestoreDataManager.this, "Downloading file for partial receipt: {}", partialReceipt.driveId);
-                    return downloadFileForReceipt(partialReceipt, downloadLocation);
+                    Single<Optional<java.io.File>> singleOptional = downloadFileForReceipt(partialReceipt, downloadLocation);
+                    Optional<java.io.File> optionalFile = singleOptional.blockingGet();
+                    if (!optionalFile.isPresent()) {
+                        missingFileCount.getAndIncrement();
+                        if (missingFileCount.get() > ALLOWED_DOWNLOAD_FAILURES) {
+                            return Single.error(new Exception("Aborting import: More than five missing files"));
```
wbaumann: Let's create a unique exception type for this

PR #202, opened 2020-02-04
Federal mileage reimbursement updates

HeinousGames:
A user (https://smartreceipts.freshdesk.com/a/tickets/2344) reported that when he put .575 in as his distance rate, any odd mileage amounts come out to half cents, and are then rounded up. His total is ahead by 3 cents because of this. I modified the total calculation rate to allow up to 3 decimals, and then round down the total to two decimal places.


app/src/main/java/co/smartreceipts/android/model/impl/ImmutableNetPriceImpl.java
```diff
@@ -54,22 +55,22 @@ public ImmutableNetPriceImpl(@NonNull PriceCurrency baseCurrency, @NonNull List<
             final PriceCurrency currencyForPriceToAdd;
             if (price.getExchangeRate().supportsExchangeRateFor(baseCurrency)) {
                 priceToAdd = price.getPrice().multiply(price.getExchangeRate().getExchangeRate(baseCurrency))
-                        .setScale(DEFAULT_DECIMAL_PRECISION, RoundingMode.HALF_UP);
```
wbaumann: Let's update the `DEFAULT_DECIMAL_PRECISION` to 3 instead

And where to we set the scale back to `2`?

HeinousGames: lines 72 and 73 we set the final total and possibly incorrect total price's scales to the default which is currently two. if i change the default to three, should i make a new default static for just this class?

app/src/main/java/co/smartreceipts/android/model/impl/ImmutableNetPriceImpl.java
```diff
@@ -55,22 +54,22 @@ public ImmutableNetPriceImpl(@NonNull PriceCurrency baseCurrency, @NonNull List<
...
                 currencyForPriceToAdd = price.getCurrency();
                 areAllExchangeRatesValid = false;
             }
             possiblyIncorrectTotalPrice = possiblyIncorrectTotalPrice.add(priceToAdd);
             final BigDecimal priceForCurrency = currencyToPriceMap.containsKey(currencyForPriceToAdd) ? currencyToPriceMap.get(currencyForPriceToAdd).add(priceToAdd) : priceToAdd;
             currencyToPriceMap.put(currencyForPriceToAdd, priceForCurrency);
         }
-        this.totalPrice = totalPrice.setScale(DEFAULT_DECIMAL_PRECISION, RoundingMode.HALF_DOWN);
-        this.possiblyIncorrectTotalPrice = possiblyIncorrectTotalPrice.setScale(DEFAULT_DECIMAL_PRECISION, RoundingMode.HALF_DOWN);
+        this.totalPrice = totalPrice.setScale(TOTAL_DECIMAL_PRECISION, RoundingMode.HALF_DOWN);
```
wbaumann: At some point in the future, we should look into JodaMoney or JSR 354 for this (instead of our custom implementation), since that'll help make things easier :)

One risk here (not the end of the world given our current user base is mostly in the US) is that it doesn't handle currencies that use different precision (e.g. Japanese Yen has 0 decimal places). Not a regression or anything but something to note here :)

PR #206, opened 2020-03-11
Release 4.20.2.2281

HeinousGames:


app/build.gradle
```diff
@@ -22,8 +22,8 @@ android {
     defaultConfig {
         minSdkVersion rootProject.ext.minSdkVersion
         targetSdkVersion rootProject.ext.targetSdkVersion
-        versionCode 2201
-        versionName "4.20.2.2201"
+        versionCode 2261
+        versionName "4.20.2.2261"
```
wbaumann: Let's bump the version here to `4.22.0.xxx`, so it's clear that we're on a new one

wbaumann: Actually - let's do this:

Send one over with `4.21.0.xxx` (I'll merge and cut the release branch). Then we'll bump this version on master again

PR #208, opened 2020-03-16
Exif comment metadata

HeinousGames:


app/src/main/java/co/smartreceipts/android/workers/EmailAssistant.java
```diff
@@ -437,10 +435,23 @@ protected WriterResults doInBackground(Trip... trips) {
 
                     if (!filterOutReceipt(mPreferenceManager, receipt)) {
                         if (receipt.hasImage()) {
+
+                            final List<Column<Receipt>> csvColumns = mDB.getCSVTable().get().blockingGet();
+
+                            StringBuilder userCommentBuilder = new StringBuilder();
+
+                            for (Column<Receipt> col : csvColumns) {
```
wbaumann: What's the purpose of this part? Generating the comment string to save to the EXIF?

HeinousGames: Yes. I was using just a string with += for all of my concatenation but the IDE warned about how I should use a StringBuilder and updated it for me. The whole description is "ColumnName: ColumnValue\n" for each column in their csv settings. I noticed that some exif readers can read all of the data, but some can't or show only some of the image description.

PR #209, opened 2020-03-18
[FLOSS] Removed Crashlytics and Google Drive dependencies for floss flavor

JuliaSoboleva:
In this PR:
* added `automatic_backups` module that has `floss` and `notFloss` flavors because of built-in google proguard rules. I also hid automatic backups section from `BackupsFragment` for floss flavor.
* added  `oss_licenses` module with `floss` and `notFloss` to isolate `com.google.android.gms:play-services-oss-licenses` dependency

As a result, now we have separated modules for features that need proprietary dependencies:

1. `analytics` module: has `floss` and `notFloss` flavors. `notFloss` depends on:
        * `com.google.firebase:firebase-core`
        * `com.crashlytics.sdk.android:crashlytics`
2.  `automatic_backups`: has `floss` and `notFloss` flavors. `notFloss` depends on:
        * `com.google.android.gms:play-services-auth`
        * `com.google.apis:google-api-services-drive`
        * `com.google.api-client:google-api-client-android`
        * `com.google.http-client:google-http-client-gson`
3. `aws`: has no flavors, depends on:
        * `com.amazonaws:aws-android-sdk-core`
        * `com.amazonaws:aws-android-sdk-s3`
4. `oss_licenses`: has `floss` and `notFloss` flavors. `notFloss` depends on:
        * `com.google.android.gms:play-services-oss-licenses`
5. `push`: has `floss` and `notFloss` flavors. `notFloss` depends on:
        * `com.google.firebase:firebase-core`
        * `com.google.firebase:firebase-crash`
        * `com.google.firebase:firebase-messaging`

I think there are no more inappropriate dependencies in the app.

analyze_floss_apk

As you can see, eventually, 4/5 new modules have `floss` and `notFloss` flavors that allow us to not include proprietary dependencies code at all. 

Currently, just one `aws` module contains `amazonaws` dependencies, but in release floss flavor it's cut out by the Proguard, so the final release apk file doesn't contain any Amazon code. 

I'm not sure if it's okay for F-Droid, and I can change code a bit and add flavors to the `aws` module similar to the other modules to make F-Droid visual inspection of the source more clear. What do you think about it?


wbaumann:
Hey sorry for the delayed review. Thanks for taking the time to put this together! I think this should be good:

> I'm not sure if it's okay for F-Droid, and I can change code a bit and add flavors to the aws module similar to the other modules to make F-Droid visual inspection of the source more clear. What do you think about it?

But I'll ping their team to find out. Great work and stay safe :)

wbaumann:
CC @HeinousGames on this one. Lots of changes here, so could you do a test run or two on your end and make sure no issues emerge :)

PR #211, opened 2020-03-28
Editing receipt shows auto complete

HeinousGames:
These changes apply to all editable fields (comment, location, and name inside trip, receipt, and distance). Is that okay?


wbaumann:
@HeinousGames - Looks good but the build is failing

PR #218, opened 2020-05-11
Second tax feature

JuliaSoboleva:
* Updated tax fields appearance and tax autocomplete list appearance
* Fixed small bug with autocomplete tax values (it didn't work with different decimal separators except `.`)
* Added additional settings: "Include Second Tax Field" checkbox, "Default Second Tax Percentage", "Second Tax Name" for tax2 and "Tax Name" for tax1
* I also added interactive moment to the tax settings, so now if "Include Tax Field" is not checked, then all other tax settings are not available. And analogically if "Include Second Tax Field" is turned off then all tax2 settings are not available
* Updated DB version to support `TAX2` column in the `ReceiptsTable`
* Updated report columns to support tax2
* Added new separate report columns for Tax1 and Tax2 that will use custom names as headers
* Added new informational tooltip "Tap to configure the second tax field" that leads to the receipt settings


wbaumann:
Thanks @JuliaSoboleva !

PR #222, opened 2020-06-08
Smr size bug

HeinousGames:
I fixed the bug where the user importing a backup and then making a new one sees their backup double in size. Then I discovered two other bugs and attempted fixes for them as well. The other two bugs involve importing multiple backup files from the automatic or manual sections. If a user chooses to overwrite the import, we see the old receipts deleted from the app, but they still remain in external storage (same how the zip file got stored). So if the user later makes a new backup file, although the app shows the old receipts were deleted, the new backup file will still contain the old deleted receipts because they weren't cleared from external storage. [I added logic on the manual side to delete the old receipts and I think I put it in a good spot.](https://github.com/wbaumann/SmartReceiptsLibrary/commit/8153c5070b719f152f26e067419aca7f1ec86c58)

The same bug could happen after importing an automatic backup, but I don't feel totally comfortable with the same solution for automatic backups being imported. The problem is that the automatic import could fail while trying to download the new files. [The only way to make sure the old files are gone from external storage is to delete them before trying the import.](https://github.com/wbaumann/SmartReceiptsLibrary/commit/032a486727b80bf031c9b16cd5ad54331dab7d2b) This could result in us deleting the user's receipts right before the import fails, resulting in the user not having current receipts or the receipts from the backup file. If I move the logic to after we know we have the files, we end up deleting the ones that were just imported. Should I remove the external directory clearing logic for auto backups before this code gets merged in?


wbaumann:
Hey @HeinousGames -

For this:

> The same bug could happen after importing an automatic backup, but I don't feel totally comfortable with the same solution for automatic backups being imported. The problem is that the automatic import could fail while trying to download the new files. The only way to make sure the old files are gone from external storage is to delete them before trying the import. This could result in us deleting the user's receipts right before the import fails, resulting in the user not having current receipts or the receipts from the backup file. If I move the logic to after we know we have the files, we end up deleting the ones that were just imported. Should I remove the external directory clearing logic for auto backups before this code gets merged in?

Yea - I think we should skip this. Seems risky and could cause us to accidentally delete all their data. Let's remove this piece :)

app/src/main/java/co/smartreceipts/android/sync/manual/ManualRestoreTask.java
```diff
@@ -158,6 +156,14 @@ private Completable restoreDataToSingle(@NonNull final Uri uri, final boolean ov
             final SDCardFileManager external = persistenceManager.getExternalStorageManager();
             StorageManager internal = persistenceManager.getInternalStorageManager();
 
+            // delete old receipt files from the app so they aren't included in future backups
+            if (overwrite) {
+                Logger.info(ManualRestoreTask.this, "Deleting old files from external storage");
```
wbaumann, line 161: This seems quite risky too. Is there a way to make sure we're only deleting the unlinked files? I worry that we're deleting everything, and this could accidentally cause data loss

HeinousGames: I guess we could check what files the backup contains and do a comparison against the final external directory after we know we successfully imported the backup. Then we only delete directory files and the receipts images/pdfs in those directories, and only if the user is choosing to overwrite. I believe we could use the same logic for the automatic backup too. I've added a commit with both sections removed. Let me know if you'd like me to attempt such a task.

wbaumann: I think it's good. To confirm with the latest change, all we now do is delete the one ZIP for the manual imports, correct?

HeinousGames: Yes. The current PR change is to specifically delete the ZIP that gets added to the external storage after it's extracted. I also updated a lot of relative class files to remove warnings (such as removing modifiers from methods that imply the same behavior).

PR #225, opened 2020-06-20
Crashlytics updates

HeinousGames:
I realized that the crashlytics update guide from Google didn't mention initializing FirebaseApp which was causing the app to crash when attempting to log a crash. I fixed this in this PR.

I also noticed in the old code, when initializing Crashlytics we only did it if we were in production app or if the user turned off crashlytics. I'm not sure if this was an error, here is what the old code looked like in the CrashReporter.kt class: https://github.com/wbaumann/SmartReceiptsLibrary/pull/223/commits/9dc89a1ff813169df4e047acef8c3a7e88554031

In this PR I updated it to check if we are in debug first, and then use the direct boolean value to enable crashlytics or not. Is this okay?


analytics/src/notFloss/java/co/smartreceipts/analytics/crash/CrashReporter.kt
```diff
@@ -10,6 +11,10 @@ class CrashReporter @Inject constructor() : CrashReporterInterface {
      */
     override fun initialize(isCrashTrackingEnabled: Boolean) {
         // Set up Crashlytics, disabling it when the user has elected to disable the functionality
-        FirebaseCrashlytics.getInstance().setCrashlyticsCollectionEnabled(isCrashTrackingEnabled)
+        if (!BuildConfig.DEBUG) {
```
wbaumann, line 14: Can you double check that this works? There was a "bug" a while back in which `BuildConfig.DEBUG` was always set to `false` in library modules (i.e. outside the `app`).

app/src/main/java/co/smartreceipts/android/SmartReceiptsApplication.java
```diff
@@ -183,6 +184,8 @@ public DateFormatter getDateFormatter() {
 
     private void init() {
 
+        FirebaseApp.initializeApp(this);
```
wbaumann, line 187: I think we need to pull this out into the `nonFloss` applications, since it won't be available for the `floss` builds

analytics/src/notFloss/java/co/smartreceipts/analytics/crash/CrashReporter.kt
HeinousGames: When I debug and step through it goes to the correct section.

wbaumann: Cool that works. I wasn't sure

PR #233, opened 2020-07-31
Feature thousand separator

JuliaSoboleva:
* Removed old price legacy classes
There were some tricks for cases when the currency is not set, now it's impossible and all our builder factories now set default currency value. So I was able to remove old classes like `ImmutableLegacyNetPriceImpl`, `ImmutablePriceImpl`, `ImmutableNetPriceImpl`, `LegacyTripPriceImpl`.

* Currencies
We've had `PriceCurrency` class before but Joda Money has their `CurrencyUnit` which also contains information about currency decimal places, so I used their class. Different currencies have different decimal places values (for ex. UYI currency has 0 decimal places) and now this moment is taken into account.
Since we have a well-defined list of supported currencies and there're some non ISO4217 ones I created `CurrencyInitializer` class that registers all currencies that we need to support but that are not included to the library by default.
There was a nuance with `DOGE`, `SC`, and `SJCX` currencies because Joda Money defines that supported currency codes must contain just 3 letters. I decided to remove the support of these three currencies due to their rarity.

* Now Joda Money deals with price formatting
We use their `MoneyFormatter` class that is locale-based and uses proper decimal and grouping separator symbols.
I also added new `PriceInputEditText` for `ReceiptCreateEditFragment` and `DistanceCreateEditFragment`. This editText allows user to input just proper symbols (digits + locale-based decimal separator), formats price with defined decimal places per request (I set it to format price on focus lose) and solves one old Android problem (when decimal separator is not '.' Android allows more than one separator to be entered).


wbaumann:
Looks good @JuliaSoboleva . Before merging, are there any areas that I should do a deeper review on? I know it's more sensitive with prices and money, so I wanted to double check there. Thanks!

JuliaSoboleva:
@wbaumann I think that the most sensitive parts are rounding and decimal places.


**1) Prices rounding mode**

JodaMoney provides two basic classes: `Money` with the standard decimal places defined by the currency and `BigMoney` with unrestricted decimal place precision.

So I used the `Money` class for `SinglePriceImpl` that is pretty clear I think. It takes `BigDecimal` and `CurrencyUnit` and forms from this `Money` with `RoundingMode.HALF_UP`. The rounding mode is needed because of defined decimal places. For example, when the user inputs $5.658 while receipt creation we'are converting it to $5.66.

`MultiplePriceImpl` is more interesting. I used `BigMoney` inside, that allows us to accumulate total value without loss of accuracy. But we still need to calculate formatted strings based on currency decimal places (decimal, currency, and currency code formatted) so I copied previous code where `RoundingMode.HALF_DOWN` was used.

Maybe using of same `RoundingMode.HALF_DOWN` or `RoundingMode.HALF_UP` for all kinds of prices would be a better solution?



**2) Distance's rate decimal places**

The next disputable thing is distance rate precision.

Previously this code was used in `DistanceBuilderFactory`

```
final int precision = ModelUtils.getDecimalFormattedValue(distance.multiply(rate), Distance.RATE_PRECISION).endsWith("0") ? Price.TOTAL_DECIMAL_PRECISION : Distance.RATE_PRECISION;
```

where `Price.TOTAL_DECIMAL_PRECISION = 2` and `Distance.RATE_PRECISION = 3`

I've set `RATE_PRECISION` as default but it needs improvement. 

I'm not sure is it a real situation when the rate has more decimal places than currency allows? And how it should look like when for example the UYI currency is used (it has 0 decimal places)?

wbaumann:
I think for this:

> Maybe using of same RoundingMode.HALF_DOWN or RoundingMode.HALF_UP for all kinds of prices would be a better solution?

Let's stick with one way. Do you know what `BigMoney` uses by default? I think we should aim to match that.

Second - do we have any tests that test some of these rounding edge cases? I'd be interested to test this out "before" and after to see if we maintain the same behavior in both.

Code with, everything is 👍. I just wanted to double check that we don't break the money side :)

JuliaSoboleva:
Hey @wbaumann :)

JodaMoney doesn't use any rounding mode as default one. They use `BigDecimal.ROUND_UNNECESSARY` that leads to `ArithmeticException` when the rounding is necessary.

I found [here](http://www.javapractices.com/topic/TopicAction.do?Id=13) that 
> the ROUND_HALF_EVEN style of rounding introduces the least bias. It is also called bankers' rounding, or round-to-even.

Are you okay with using `RoundingMode.HALF_EVEN` for our price classes?

As for the distance's rate, yes, we had some tests in `DistanceTest`, I didn't change it. That tests check that decimal formatted rate has 3 decimal places when currency formatted rate has 2 decimal places. So, logically all works like before, but visually users now will see something like "5.000" in the `DistanceCreateEditFragment` (`RATE_PRECISION` constant was set to 3 before my changes, but now formatting is more strict).

wbaumann:
Cool all sounds good there. And yes - `RoundingMode.HALF_EVEN ` works for me. Is that already all in place? Or anything else to merge there? Thanks!

JuliaSoboleva:
@wbaumann Now it's all in place :)

PR #234, opened 2020-08-09
Remove tax column

HeinousGames:


app/src/main/java/co/smartreceipts/android/model/impl/columns/categories/CategoryColumnDefinitions.kt
```diff
@@ -57,9 +59,12 @@ class CategoryColumnDefinitions(private val reportResourcesManager: ReportResour
 
         for (definition in actualDefinitions) {
             // don't include PRICE_EXCHANGED definition if all receipts have same currency
-            if (!(definition == PRICE_EXCHANGED && !multiCurrency)) {
-                columns.add(getColumnFromClass(definition))
+            // don't include TAX definition if tax field is disabled
+            if ((definition == PRICE_EXCHANGED && !multiCurrency)
+                    || (definition == TAX && !taxEnabled)) {
```
wbaumann, line 64: Let's split this into two if statements for easier readability like:

```
if (!(definition == PRICE_EXCHANGED && !multiCurrency)) {
  // don't include PRICE_EXCHANGED definition if all receipts have same currency
} else if (definition == TAX && !taxEnabled)) {
 // don't include TAX definition if tax field is disabled
} else {
   columns.add(getColumnFromClass(definition))
} 
```

wbaumann:
Looks good. One nit-picky request

PR #236, opened 2020-08-10
Update total reimbursable logic

HeinousGames:


app/src/main/java/co/smartreceipts/android/workers/reports/pdf/pdfbox/PdfBoxReceiptsTablePdfSection.java
```diff
@@ -260,7 +260,9 @@ public void writeSection(@NonNull PDDocument doc, @NonNull PdfBoxWriter writer)
                 pdfBoxContext.getFontManager().getFont(PdfFontStyle.DefaultBold))));
 
         // Print the grand total (reimbursable)
-        if (!preferenceManager.get(UserPreference.Receipts.OnlyIncludeReimbursable) && !data.getGrandTotalPrice().equals(data.getReimbursableGrandTotalPrice())) {
+        if (!preferenceManager.get(UserPreference.Receipts.OnlyIncludeReimbursable)
+                && !data.getGrandTotalPrice().equals(data.getReimbursableGrandTotalPrice())
+                && !data.getReimbursableGrandTotalPrice().getDecimalFormattedPrice().equals("0.00")) {
```
wbaumann, line 265: I think we should check if it equals `BigDecimal.ZERO` instead. 

Here's why:

The formatting for the US is `0.00` but for the EU, it'll look like `0,00` instead. Checking against `ZERO`

wbaumann:
@HeinousGames - I'm noticing that the builds don't seem to be triggered on these PRs to validate that all tests pass. Any idea if something changed or broke that? Or maybe I'm just missing it in GitHub's new UI :)

wbaumann:
Ignore me. The UI is just different now

PR #238, opened 2020-08-20
Norwegian translation

HeinousGames:


wbaumann:
You worked with that one ticket on this right? Should be all good I assume?

HeinousGames:
> You worked with that one ticket on this right? Should be all good I assume?

Yup! She translated the files and I just added them to the locale's folders. Changing my device to the language made the UI change, and I was able to successfully choose Norwegian for my report and have it translated.

PR #239, opened 2020-09-03
Google min api

HeinousGames:


app/src/main/java/co/smartreceipts/android/distance/editor/DistanceCreateEditFragment.kt
```diff
@@ -89,7 +90,7 @@ class DistanceCreateEditFragment : WBFragment(), DistanceCreateEditView, View.On
 
     private lateinit var paymentMethodsAdapter: FooterButtonArrayAdapter<PaymentMethod>
 
-    private lateinit var itemToRemoveOrReAdd: AutoCompleteResult<Distance>
```
wbaumann, line 92: How come we got rid of all these `lateinit`?

HeinousGames: Updating to api 29 resulted in the compiler throwing errors for adapter.getItem. GetItem could return a null, so the actual variable has to be initialized as a null, resulting in it not allowed to be a lateinit, as it is being initialized now.

wbaumann: If it can be `null`, we should remove the `!!` syntax then. Let's do this type of approach:

```kotlin
itemToRemoveOrReAdd?.let {
  it.whatever()
}
```

wbaumann: That was we can maintain null safety with the new changes

HeinousGames: It really can't be null. What's happening (in my understanding) is that the adapter can return a null item (even though we only call it to get an item). So itemToRemoveOrAdd is never actually null, it just has to be initialized as such because of the getAdapter method having the option of returning a nullable. Turning itemToRemoveOrAdd to a nullable refactors the methods that use that item, but we can enforce that it's non null because the user can't remove a non-item from the suggested drop down list.

wbaumann: Sounds good

PR #241, opened 2020-09-07
Show receipt total

HeinousGames:
user request to show the number of receipts in a trip. this change allows you to see the receipt amount in the row in the trip list. the user may want further visibility in seeing the receipt amount when inside the report. we could possibly add that value on the graphs page, or update the Receipts tab to show the number of receipts (below the action bar).


app/src/main/java/co/smartreceipts/android/adapters/TripCardAdapter.java
```diff
@@ -46,4 +46,9 @@ protected void setDateTextView(TextView textView, Trip data) {
 		textView.setText(getContext().getString(R.string.trip_adapter_list_item_to, start, end));
 	}
 
+	@Override
+	protected void setCategory(TextView textView, Trip data) {
+		textView.setText("Receipts: " + data.getSize());
```
wbaumann, line 51: This should come from a `R.string` resource

app/src/main/java/co/smartreceipts/android/model/Trip.kt
```diff
@@ -51,9 +51,10 @@ class Trip @JvmOverloads constructor(
     override var price: Price = PriceBuilderFactory().setCurrency(tripCurrency).build(),
     /**
      * The daily sub-total [Price] (i.e. all expenditures that occurred today) for this trip,
-     * daily sub-total of a trip exists as a function of it's receipt children (and not itself)
+     * daily sub-total of a trip exists as a function of its receipt children (and not itself)
      */
     var dailySubTotal: Price = PriceBuilderFactory().setCurrency(tripCurrency).build(),
+    var size: Int,
```
wbaumann, line 57: What's this for? Why not just use the `receipts.size()` here instead?

wbaumann (changes requested):

app/src/main/java/co/smartreceipts/android/model/Trip.kt
HeinousGames: We are trying to show the amount of receipts in a report, not the total price of the receipts combined. The TripFragment, where we are showing the receipt amount, has not opened a trip yet so we can't call receipts.size on it. Instead I updated the DatabaseHelper class to call receipts.size, and then store it inside the trip model to be referenced from the TripFragment.

wbaumann: Could you send a screenshot of how this works? I'm thinking about it a bit more and think it might end up being a weird experience that is sort of confusing to the larger user community, so I may want to pause on it

HeinousGames: Sure Will! Here's a quick explanation:

A user requested being able to see the number of receipts from each month for reference. I noticed that the TripAdapter utilizes the same CardAdapter that is used for Receipt and Distance lists. The layout had an additional textfield with the ID of category that was being used for the receipt and distance row but not the trip row. Since it was already built-in, I just decided to throw the receipt count there. This is what it looks like:


Screenshot_20200911-154430

Since we are on the trip fragment, we can't call receipts.size as we call for the receipts after opening a trip to display them. We can hold off on adding it if you think I should save the category field for some other form of data, or if we need iOS to match before merging.

wbaumann: Yea - I think visually, you did a good job here. I juts thing we should hold off, since it'll open us up to a bunch of other custom fields that add extra noise here.

IMO - We add this to the "graphs" section instead, so can use that as a clearer "data" section

PR #242, opened 2020-09-10
Fixes for upgradeDatabase() espresso tests

JuliaSoboleva:
Fixed 2 espresso tests

The first was failing because `getDecimalFormattedExchangeRate()` result now contains a grouping separator symbol, and I just fixed the expected string value.

The second test was failing because `ReceiptDatabaseAdapter` was not ready to read price values with comma decimal separator. Such values shouldn't be present actually, but they can be found in fairly old versions of the database.


wbaumann:
Thanks @JuliaSoboleva . CC @HeinousGames :)

PR #256, opened 2020-12-30
Removed "smart" functions from price input editText for app stability

JuliaSoboleva:
This should fix bugs connected with price input editText.

Now price input works like it was working before grouping separators integration. It ignores thousand separators and will not format price value when the focus is changed.

Elsewhere the use of price formatting is preserved.


app/src/main/java/co/smartreceipts/android/model/utils/ModelUtils.kt
```diff
@@ -137,18 +137,7 @@ object ModelUtils {
             throw NumberFormatException("Cannot parse an empty string")
         }
 
-        // Note: for some locales grouping separator symbol is non-breaking space (code = 160),
-        // but incoming string may contain general space => need to prepare such string before parsing
-        val groupingSeparator = decimalFormat.decimalFormatSymbols.groupingSeparator.toString()
-        val nonBreakingSpace = ("\u00A0").toString()
```
wbaumann, line 143: Do we need to maintain this behavior?

app/src/main/java/co/smartreceipts/android/model/utils/ModelUtils.kt
```diff
@@ -137,18 +137,7 @@ object ModelUtils {
...
-
-        val parsedNumber =
-            when (groupingSeparator) {
-                nonBreakingSpace -> decimalFormat.parse(number.replace(" ", nonBreakingSpace))
-                else -> decimalFormat.parse(number)
-            }
-
-        return BigDecimal(parsedNumber.toString())
+        return BigDecimal(number.replace(",", "."))
```
wbaumann, line 140: What happens here if someone passes an input string like `1,000.00`? Can we protect this case?

app/src/main/java/co/smartreceipts/android/model/utils/ModelUtils.kt
JuliaSoboleva: All uses of this method come from the user's input, and `android:inputType="numberDecimal|numberSigned"` makes these checks meaningless

app/src/main/java/co/smartreceipts/android/model/utils/ModelUtils.kt
JuliaSoboleva: Android system will handle this case for us. `EditText` with `android:inputType="numberDecimal|numberSigned"` won't allow user to input any other character than `.`. Even if a user copies from somewhere "1,000.00" after pasting he will see "1000.00" in the input field.

This approach is not flexible for different locale-dependent decimal separators but it's much more stable. So, I propose to use it for price inputs while in all other places where price value is not editable we will show to the user a well-formatted string with correct decimal and thousand separators.

PR #276, opened 2022-07-19
New Subscriptions Model

JuliaSoboleva:
Current state:

`Standard subscription` = free OCR
`Premium Subscription` = free OCR + backups + no ads (and all the features from Smart Receipts Plus)

Texts on subscriptions activity (on design in Figma actually) need to be updated and translated. The description of the subscriptions on this screen is not correct at the moment.

New Subscriptions turns on by `ConfigurableResourceFeature.SubscriptionModel` bool flag (now set to `false`)

If new Subscription Model is used than there are some changes for main menu:
- add new "Subscriptions" option that opens `SubscriptionsActivity`
- remove option "Smart Receipts Plus"
- remove option "OCR Scans" _(That is pretty controversial decision to me because when we will move to new Subscriptions model we will have users with some amount of bought OCR scans. And by hiding OCR menu we just take away any control over it from them. The situation is also complicated by the fact that now we can't determine whether the user bought the scans or received them for free.)_

No Analytics and A/B tests for Subscriptions

Subscription purchase process:
- not logged in users can't buy subscription, they will be redirected to the login screen
- when purchase is completed we try to upload it to the backend first
- only if uploading was successful we acknowledge that purchase to Google and provide functionality to the user
- if uploading fails we will retry all the process of uploading and acknowledgement on start of the app while Google will keep that purchase unacknowledged (about 3 days)

Place of possible problems: the backend checks subscription status every 3 days, so based on the server response, we can allocate benefits to user for a couple of days longer than they should.


app/src/androidTest/java/co/smartreceipts/android/test/espresso/BaseEspressoTests.kt
```diff
@@ -129,7 +129,7 @@ class BaseEspressoTests {
         onView(withId(R.id.action_save)).perform(click())
 
         // Wait until everything loads
-        onView(isRoot()).perform(waitForView(R.id.bottom_app_bar, 20000))
+//        onView(isRoot()).perform(waitForView(R.id.bottom_app_bar, 20000))
```
ruslangurbanzade, line 132: need to be removed.

app/src/flossFlavor/AndroidManifest.xml
```diff
@@ -23,6 +23,9 @@
             android:theme="@style/Theme.SmartReceipts.OSS"
             tools:node="remove" />
 
+        <activity
+            android:name="co.smartreceipts.android.subscriptions.SubscriptionsActivity"
```
ruslangurbanzade, line 27: Creating new activity for new features is not good solution. Based on google best practice using a single activity can reduce memory usage.

PR #287, opened 2023-07-13
Fix crashlytics main bugs

qwaskhaled:


app/src/main/java/co/smartreceipts/android/search/viewholders/TripViewHolder.kt
```diff
@@ -0,0 +1,59 @@
...
+        binding.textName.setTextColor(colorDefault)
+        binding.textPrice.setTextColor(colorDefault)
+        binding.textPrice.text = item.price.currencyFormattedPrice
+        binding.textName.text = item.name
+
+        binding.textDetails.text =
+            itemView.context.getString(R.string.trip_adapter_list_item_to, start, end)
+        binding.textDetails.setTypeface(null, Typeface.NORMAL)
+        binding.textDetails.setTextColor(colorText)
+        binding.imageSyncState.visibility = View.GONE
```
EvsenevDev, line 49: It's always `GONE`, should we show sync status here?

qwaskhaled: I think this xml is reused more than once. In some views imageSyncState can be visible.

EvsenevDev: It's ok, but Trips can show correct sync state, here is always hidden?

